Issue #246, opened 2023-03-28
Commands not properly constructed in some cases


carueda:
**Describe the bug**
Commands not properly constructed in some cases

**To Reproduce**
1. Try to issue a `failComponent` command to some vehicle (say `sim`)
4. Error is displayed at the last step.

**Expected behavior**
No error, just the command successfully issued.

**Additional context**

The request is reflected like this in the backend:
```
2023-03-28 15:06:15,507 (http-nio-8080-exec-19) DEBUG [o.m.t.r.s.CommandRouteHandler] - sendCommands:
  vehicle=sim
  commandText=sched asap ""
  commandNote=
  schedDate=asap
  schedId=2b6qf (server generated)
  runCommand=false
  email=carueda@mbari.org
  destinationAddress=null
```

It is `commandText=sched asap ""` piece that clearly indicates the command was not properly formed.

jimjeffers:
@carueda I was able to replicate the issue. It looks like the default selection is not carrying over in the wizard state. If you choose an alternative template of the command and then select the default failComponent template the proper `sched asap` gets sent to the API. We just need to fix the default state on the command step to ensure it's carried over to the next step in the wizard.

<img width="1026" alt="CleanShot 2023-03-28 at 15 45 15@2x" src="https://user-images.githubusercontent.com/9150/228383601-15e90bfd-1856-4b98-8fc0-a98dd3ba440b.png">

jimjeffers:
@ksalamy I believe this is the line you're going to be looking for:
https://github.com/mbari-org/dash5/blob/0b527bff294ea6e20a4070bf4612198680a57503/packages/react-ui/src/Modals/CommandModal.tsx#L178

I think setting the initial state to `syntaxVariations[0]` instead of `null` might solve this. Let me know if you get blocked on this and I can dig further if need be.

jimjeffers:
...or to be null safe 

```ts
useState<string | null>(syntaxVariations?.[0] ?? null)
```

heavyB:
Note that dash 5 is using the schedule asap "<string>" command incorrectly. Help schedule defines appropriate use. However, just simply send "failc" rather than "schedule asap "failc"" for any asap commands. There's no need to invoke the scheduler. Please see dash 4 for appropriate syntax. 

catalina2>sched asap "failc" 
2023-04-11T22:18:45.982Z,1681251525.982 [CommandExec](IMPORTANT): got command schedule asap "failc" 
[...]

PR #280, opened 2023-03-30
Resolve Issue 246 - Command Not Constructed

jimjeffers:
@ksalamy please see this fork. I found that we were not assigning the default selected syntax in the primary component for the  `CommandModal`. Additionally when I initially implemented this component it seemed logical for our team to assign the generated argument for the selected syntax to the ref that tracks changes. This turned out to be the primary cause of this bug because the side effect that listens for changes to the generated command needs to run in order to trigger the necessary state update in the `CommandModal` component which tracks the state for every step of the wizard process.

Also added an additional unit test for the `CommandModal` which should hopefully prevent any regression from this happening in the future. This new complete test can be duplicated in case future edge case bugs are identified in the wizard workflow.

### DIFF
--- a/packages/react-ui/src/Modals/CommandModal.sampleProps.tsx
+++ b/packages/react-ui/src/Modals/CommandModal.sampleProps.tsx
@@ -1,0 +1,139 @@
+import { CommandModalProps } from './CommandModal'
+
+export const syntaxVariations: CommandModalProps['syntaxVariations'] = [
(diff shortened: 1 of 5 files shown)


jimjeffers:
@ksalamy could you please review this PR and let me know if you have any questions? This should close #246 .

ksalamy:
Hi Jim,

Looks good.  No merge issues. Performed a quick test on Daphne with success!

PR #283, opened 2023-04-18
maxZoom Fixes

jimjeffers:
This adjusts the maxZoom level for the `DeploymentMap` component and also removes a hardcoded value on one of the tile layers in the general `Map`component.


ksalamy:
Thanks Jim.

PR #294, opened 2023-06-29
Depth coord now loaded via google elevation service.

jimjeffers:
Implemented onDepthRequest on the [...deployment].tsx page. Still need to confirm we're able to extract the google API key from the site config. If not we should probably use an environment variable to embed the maps API key more reliably.

Also the API key must have the google elevation service enabled in order for requests to successfully be made.

CleanShot 2023-06-28 at 18 16 30


ksalamy (approved):
Updates look good. Will test out.

PR #300, opened 2023-08-21
Replace font-awesome-pro w/ free version.

jimjeffers:
I downgraded to the free version of font-awesome as my font-awesome license is having bandwidth issues and we can no longer build with my key. MBARI could purchase a pro license but I suspect it's not necessary as nearly all of the same icons are still available in the free version we just have less options on whether or not they are solid or light weight. Let me know if this will be acceptable.


ksalamy:
Hi Jim,  Honestly I do not think this will be a huge issue so long as we can still access the same icons.  Thanks.

Issue #312, opened 2024-02-16
PIC/On-call not displayed


sqshemet:
**Describe the bug**
The PIC/On-call history is correct, but the dash shows no PIC/no On-Call and Unassigned/Unassigned

**To Reproduce**
1.Look at PIC/On-Call during any active deployment

**Expected behavior**
PIC and On-Call should show users currently signed in

**Screenshots**
![image](https://github.com/mbari-org/dash5/assets/1900768/4aec85e2-28ce-49b1-8f16-04344120cfe4)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Ubuntu 22.04
 - Browser [e.g. chrome, safari] Firefox
 - Version [e.g. 22] 122.0

beroe:
Also seeing this... +1

heavyB:
UX front end needs menu button to support this.

jimjeffers:
## Analysis of PIC/On-call Display Issue

I've investigated issue #312 where the PIC/On-call information is not displaying correctly in the dashboard. Here's what I found:

### Data Flow
1. The data is fetched using `usePicAndOnCall` hook in `/apps/lrauv-dash2/pages/vehicle/[...deployment].tsx` (line ~86)
2. This hook correctly fetches PIC/On-call data from the API
3. The data is then passed to the `OverviewToolbar` component (lines 200-203):
   ```tsx
   <OverviewToolbar
     vehicleName={vehicleName}
     pilotInCharge={picAndOnCall?.[0].pic?.user}
     pilotOnCall={picAndOnCall?.[0].onCall?.user}
     deployment={...}
   />
   ```

### Problem Areas
Based on the issue description and code analysis, there are three likely sources of the problem:

1. **API Response Handling**: 
   - The API might be returning data in a format that doesn't match what the front-end expects
   - Check for discrepancies between the expected response shape in `getPicAndOnCall.ts` and the actual API response

2. **Data Transformation/Presentation**:
   - The `usePicAndOnCall` hook (in `/packages/api-client/src/react-query/User/usePicAndOnCall.ts`) correctly fetches data but values may be coming back as `undefined` or `null` when API response doesn't include expected fields
   - When data comes back with no PIC or On-call, the UI shows "no PIC/no On-Call" instead of handling this case properly

3. **Toolbar Component**:
   - The `OverviewToolbar` component needs to better handle edge cases when no PIC/On-call is actually assigned

### Suggested Investigation Steps
[...]

PR #355, opened 2025-03-04
Incorrect PIC / OnCall displayed

ZackLyon:
References #312 

This issue was likely due to a type error caused by native checkbox behavior, where a string is output instead of an array when only one checkbox is selected. This was causing the PIC/OnCall not to be updated properly in some cases. Additionally, the PIC / OnCall was not updating in the UI without refreshing. Both of these issues have been updated.
.env was also added to the gitignore list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ yarn-error.log*
 .DS_Store
 
 # local env files
+.env
 .env.local
 .env.development.local
 .env.test.local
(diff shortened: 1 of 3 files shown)


ZackLyon:
Additional note: PIC and OnCall are only displayed if an authorized user is logged in. Let me know if we want to change this.

jimjeffers (approved):
These changes look fine! I saw you had questions about whether or not the API expected an array or not? Per our conversation earlier today there are API docs. I'm requesting access to them as I went digging and got a 404. If you were able to successfully send the request to the API and the PIC persisted on the SIM then your update is correct and we can merge. Let me know if you were able to confirm that the API returns a 20x response for you at your leisure and we'll merge this. Thanks @ZackLyon

ZackLyon:
@jimjeffers - Here's the 200 responses when changing PIC / OnCall; the changes also persist on the UI now. I'm attaching a screenshot of the updated warning that pops up when a vehicle is not selected when interacting with the role reassignment modal in case there are any notes about that.

#### Successful reassignment updates
Screenshot 2025-03-04 at 10 53 38 AM

#### Updated warning message when no vehicle is selected
Screenshot 2025-03-04 at 10 54 08 AM

PR #359, opened 2025-03-06
Automatically redirect root page to dash page

ZackLyon:
Going to the root page now redirects to the dash page when working locally instead of giving a 404.

Redirect applied to localhost ports 3000 and 3001.


ZackLyon:
@jimjeffers - I've updated this to only redirect in development.

jimjeffers (approved):
Looks good!

PR #361, opened 2025-03-10
Update event, vehicle position and chart endpoints to use Unix epoch milliseconds

ZackLyon:
Change date props for event, vehicle position and chart endpoints to use unix epoch milliseconds instead of dates to avoid timezone conversion errors.
Create helper function to make hour, day and year offsets on dates while remaining in unix epoch millisecond format.
Update sorting of dates for lastEvent in getDeployments to sort numerically instead of lexicographically.
Remove unused `to` param from getLastDeployment.


apps/lrauv-dash2/components/VehicleAccordion.tsx
```diff
@@ -47,11 +47,11 @@ const VehicleAccordion: React.FC<VehicleAccordionProps> = ({
   activeDeployment,
   currentDeploymentId,
 }) => {
-  const { data: relatedLogs, isLoading: logsLoading } = useEvents({
```
jimjeffers, line 50: Do we want this to be commented out? If so should we delete it entirely?

packages/utils/src/getAdjustedUnixTime.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Optionally adjusts the timestamp to the end of the day.
+ *
+ * @param unixTime - Unix timestamp in milliseconds.
+ * @param offsetDays - Optional number of days to adjust the date. Positive values add days, negative values subtract.
+ * @param offsetHours - Optional number of hours to adjust the date. Positive values add hours, negative values subtract.
+ * @param offsetYears - Optional number of years to adjust the date. Positive values add years, negative values subtract.
+ * @param endOfDay - Optional boolean to adjust the time to the end of the day after applying offsets. Defaults to false.
+ * @returns The adjusted Unix timestamp in milliseconds.
+ */
+export const getAdjustedUnixTime = ({
```
jimjeffers, line 14: Since this function is doing the heavy lifting for date handling throughout the app now - we should add a simple set of unit tests just to ensure future edits won't introduce an accidental regression.

apps/lrauv-dash2/lib/useCurrentDeployment.ts
```diff
@@ -11,7 +11,6 @@ const useCurrentDeployment = () => {
   const { data: lastDeployment, isLoading } = useLastDeployment(
     {
       vehicle: vehicle as string,
-      to: new Date().toISOString(),
```
jimjeffers, line 14: Removing to here doesn't have an effect as that's what the API defaults to right? Just checking!

jimjeffers (changes requested):
@ZackLyon looks great - just requesting some simple test be added for the new date function. Also can you confirm whether or not the commented code I inquired about can be deleted?

apps/lrauv-dash2/lib/useCurrentDeployment.ts
ZackLyon: The `/deployments/last` endpoint doesn't even have a `to` param :laughing:
Screenshot 2025-03-10 at 6 13 46 PM

packages/utils/src/getAdjustedUnixTime.ts
ZackLyon: @jimjeffers - We don't have testing functionality set up in `packages/utils` currently, but I'll set it up tonight and write some tests for this.

apps/lrauv-dash2/components/VehicleAccordion.tsx
ZackLyon: I'll delete the commented out code. It's pretty simple stuff so if someone needs it, it will be easy to replicate.

PR #367, opened 2025-03-12
Resolve Map Crashing Issue

### LINKED ISSUE #366: Map crashes with 'undefined is not an object (evaluating n.latitude)'
## Description

The application crashes when attempting to view a specific vehicle on the map with the following error:

```
TypeError: undefined is not an object (evaluating 'n.latitude')
(anonymous function) — 90.d16de712b20b3dde.js:1:1959
ri — framework-9115437038c14d1c.js:1:163228
Ss — framework-9115437038c14d1c.js:1:183456
(anonymous function) — framework-9115437038c14d1c.js:1:180046
S — framework-9115437038c14d1c.js:1:208258
R — framework-9115437038c14d1c.js:1:208792
```

## Source of the issue

The likely source of the bug is in the VehiclePath.tsx component when handling vehicle position data. When generating the  array around lines 153-162, there may be undefined values that aren't properly handled. The issue could be that we're trying to access  on an undefined object when constructing map coordinates.

## Potential fixes

1. Add additional null checks and fallbacks when creating the  array
2. Ensure all objects in the array being mapped have valid latitude/longitude values
3. Filter out any undefined or null values before mapping the coordinates

## Steps to reproduce

1. Navigate to a vehicle view
2. The error occurs when the map tries to render the vehicle path with certain undefined coordinate data

## Impact

This issue prevents users from viewing certain vehicles on the map, causing the application to crash.

ZackLyon:
Add additional checks for coordinate availability when determining latest vehicle position for vehicle path on map.
Resolve issue where it was determining whether to update by checking both coordinates against longitude.
Filter out bad or undefined coordinates while determining inactive vehicle path.
Add error handling to Google Maps elevation service requests.

Closes #366

### DIFF
--- a/apps/lrauv-dash2/components/VehiclePath.tsx
+++ b/apps/lrauv-dash2/components/VehiclePath.tsx
@@ -1,4 +1,3 @@
-
 import React, { useEffect, useRef, useCallback, useState, useMemo } from 'react'
 import { DateTime } from 'luxon'
 import {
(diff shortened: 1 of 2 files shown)


ZackLyon:
@jimjeffers @heavyB - Actually, I see there might be additional issues from the one I resolved before merging this.

ZackLyon:
@jimjeffers @heavyB - Ok, this should be ready for review now.

PR #370, opened 2025-03-17
Fix undefined coordinates bug

ZackLyon:
Verify latest GPS coordinates are defined before updating map center position
Compare latest GPS coordinates to center position to avoid unnecessary repeated updates


jimjeffers (approved):
Good catch @ZackLyon. I'll merge this change now.

PR #400, opened 2025-04-04
Resolved issue with PIC and On Call not being displayed for vehicles that are not deployed

ZackLyon:
- Discovered PIC / On Call endpoint does not return the users in those roles unless vehicle is deployed
- Created util function to extract PIC and On Call from event notes (based on dash4 implementation)
- Added tests for util function
- Created hook to get PIC / On Call for vehicles that are not deployed
- Implemented hook on vehicle page for handoff section and in toolbar reassignment button


jimjeffers (changes requested):
@ZackLyon looks good - here are some minor refactors to consider before merging:

## 1. Improve Regex Pattern Specificity

The current pattern is too permissive and might match unintended text. Replace:

```typescript
// Current implementation
const NOTE_REGEX =
  /.*(?:Signing|Signed|Sign) (in|off|out) as (PIC|On-Call|On Call).*/i
```

With a more specific pattern:

```typescript
// Suggested improvement
const NOTE_REGEX =
  /^(?:Signing|Signed|Sign)\s+(in|off|out)\s+as\s+(PIC|On-Call|On Call)(?:\s+|$)/i
```

This ensures:
- The statement begins with the signing action
- Requires proper spacing between words
- Doesn't match if these phrases appear in the middle of longer text

If this is too strict you can keep as is.

## 2. Create a Helper Function for Handoff Label Formatting

Extract the complex handoff label logic to a reusable function:

```typescript
// Add this helper function
const formatHandoffLabel = (
  pic: string | undefined, 
  onCall: string | undefined, 
  profileName: string
): string => {
  if (!pic && !onCall) return '';
  
  return `${shortenName(pic ?? 'Unassigned')}${
    pic === profileName ? ' (you)' : ''
  } / ${shortenName(onCall ?? 'Unassigned')}${
    onCall === profileName ? ' (you)' : ''
  }`;
};

// Then simplify the component code
const handoffLabel = loadingPic || loadingInactivePicAndOnCall
  ? '...'
  : pic || onCall 
    ? formatHandoffLabel(pic, onCall, profileName)
    : undefined;
```

## 3. Optimize Event Query with Error Handling and Pagination

We could have the `useVehiclePicAndOnCall` hook to use a shorter time span - how many events are pulled in when you query for two years? It seems like it could be a large request but if the size is not massive you can ignore this optimization:

```typescript
export const useVehiclePicAndOnCall = ({
  vehicleName,
  from,
  to,
  enabled = true,
}: UseVehiclePicAndOnCallParams): VehiclePicAndOnCallResponse => {
  // Use a more recent default time range (e.g., 3 months instead of 2 years)
  const THREE_MONTHS_AGO = getAdjustedUnixTime({
    unixTime: DateTime.now().toMillis(),
    offsetMonths: -3,
  });
  
  const { 
    data: noteEvents, 
    isLoading, 
    error 
  } = useEvents(
    {
      vehicles: [vehicleName],
      from: from ?? THREE_MONTHS_AGO,
      eventTypes: ['note'],
      to,
      limit: 1000, // Reduce from 3000 to improve performance
    },
    {
      enabled,
      retry: 2, // Add retry logic for network failures
      onError: (err) => {
        console.error('Failed to fetch PIC and On Call events:', err);
      }
    }
  );

  const picAndCall = useMemo(() => {
    if (!noteEvents?.length) {
      return { pic: null, onCall: null };
    }
    
    return extractPicAndOnCallFromNotes(noteEvents);
  }, [noteEvents]);

  return {
    ...picAndCall,
    isLoading,
    error: error ? true : false,
  };
};
```

## 4. Create a Reusable PIC/OnCall Resolver

To reduce duplication across components:

```typescript
// Add to a shared utility file
export const resolvePicAndOnCall = (
  activeData: any[] | undefined, 
  inactiveData: { pic: any; onCall: any }
) => {
  return {
    pic: activeData?.[0]?.pic?.user ?? inactiveData.pic?.user,
    onCall: activeData?.[0]?.onCall?.user ?? inactiveData.onCall?.user,
  };
};

// Then in components
const { pic, onCall } = resolvePicAndOnCall(picAndOnCall, inactivePicAndCall);
```

ZackLyon:
@jimjeffers - I reduced the time on the inactive PIC and on call request. I also reduced unnecessary replication of figuring out the PIC / on call by passing it from the deployment page to the vehicle accordion so now it's only determining it there. There were also some issues with how the inactive PIC and on call were being determined so I fixed that, too. I don't think the resolving function for the two is necessarily a good idea since their loading states are also needed. Let me know if you feel strongly about other changes that are needed.

ZackLyon:
Also, I think the stricter regex was causing issues, so I didn't use it. As they say: 'good enough for dash4, good enough for me'.

ZackLyon:
@jimjeffers - I think we might need to abandon the endpoint to get PIC and on call entirely. It isn't reliable. Here's the PIC and on call using the dash4 way of sorting through tons of notes vs the endpoint. I thought it was only unreliable for inactive vehicles, but ahi is deployed.

#### Dash5 ahi
Screenshot 2025-04-07 at 5 49 29 PM

#### Dash4 ahi
Screenshot 2025-04-07 at 5 53 48 PM

ZackLyon:
@jimjeffers - This is ready for review again. See comments above for what I did and did not change. I did replace the usePicAndOnCall hook entirely with the new one. I didn't remove the code for it though since we might use it some day if the endpoint is updated. I also updated the useVehiclePicAndOnCall hook to more closely resemble the input and output of the old hook. Now multiple vehicles can be called with it.

jimjeffers (approved):
Oof fetching 3000 notes just to provide a simple UI indication is definitely a client side work around but that's the best we can do via the front end. Looks good to me @ZackLyon. Additional tests look good too. You can merge this when ready.

Issue #409, opened 2025-04-24
Depth map crashing while trying to access uninitialized variable


ZackLyon:
**Describe the bug**
When selecting the depth map tab, the page crashes due to it trying to access a property on the depth data that is not available.

**To Reproduce**
1. Go to staging
2. Go to a vehicle deployment (select vehicle tab)
3. Click on the depth tab at the bottom of the map
4. See page crash with client side error

**Expected behavior**
Page should not crash, even if map or depth data cannot load

ZackLyon:
Closed via #410

PR #410, opened 2025-04-25
Fix depth map crashing

ZackLyon:
Fix depth map crashing due to an arrow function being used before it was defined


ZackLyon:
I just merged this since I just rearranged code for the most part. Closes #409

jimjeffers:
Nice 👍🏼

PR #417, opened 2025-05-01
Various debugging for GoogleMapApiKey services

ksalamy:
This code is building on my local development and properly loading GoogleMapsApiKey from the server but not when it is deployed on dash5-staging.mbari.org?! This PR attempts to fix various console errors on the dash5-staging code currently running as well as determine why my calls to the GoogleMapsApiKey and Elevation services are failing on the server. If there is a better option, please advise.

### DIFF
--- a/apps/lrauv-dash2/components/DeploymentMap.tsx
+++ b/apps/lrauv-dash2/components/DeploymentMap.tsx
@@ -3,7 +3,6 @@ import React, { useCallback, useState, useRef, useEffect } from 'react'
 import { useManagedWaypoints } from '@mbari/react-ui'
 import useGoogleElevator from '../lib/useGoogleElevator'
 import { VPosDetail } from '@mbari/api-client'
-// import { StationsListModal } from './StationsListModal'
 import { MapLayersListModal } from '../components/MapLayersListModal'
 import { useSelectedStations } from './SelectedStationContext'
 import { useMarkers } from './MarkerContext'
@@ -11,13 +10,17 @@ import { toast } from 'react-hot-toast'
 import { createLogger } from '@mbari/utils'
 import VehicleColorsModal from './VehicleColorsModal'
 import useTrackedVehicles from '../lib/useTrackedVehicles'
+import type { MapProps } from '@mbari/react-ui/dist/Map/Map'
 
 // This is a tricky workaround to prevent leaflet from crashing next.js
 // SSR. If we don't do this, the leaflet map will be loaded server side
 // and throw a window error.
-const Map = dynamic(() => import('@mbari/react-ui/dist/Map/Map'), {
-  ssr: false,
-})
+const Map = dynamic<CustomMapProps>(
+  () => import('@mbari/react-ui/dist/Map/Map'),
+  {
+    ssr: false,
+  }
+)
 const DraggableMarker = dynamic(() => import('./DraggableMarker'), {
   ssr: false,
 })
@@ -49,6 +52,15 @@ interface DeploymentMapProps {
   endTime?: number | null
 }
 
+// interface CustomMarkerProps
+type CustomMapProps = MapProps &
+  React.RefAttributes<L.Map> & {
+    isAddingMarkers?: boolean
+    onToggleMarkerMode?: () => void
+    trackedVehicles?: { name: string; id?: string }[]
+    mapId?: string
+  }
+
 const DeploymentMap: React.FC<DeploymentMapProps> = ({
   vehicleName,
   indicatorTime,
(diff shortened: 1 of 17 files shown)


jimjeffers:
@ksalamy @ZackLyon I'm unsure about all of these changes to get the API key loading. The existing codebase was already doing this here:

> The Google Maps API key is loaded from the Tethys API through:
> 
>   1. packages/api-client/src/axios/Info/getInfo.ts - Makes the request to /info endpoint
>   2. packages/api-client/src/react-query/Info/useSiteConfig.ts - Creates a React Query hook that calls getInfo
>   3. packages/api-client/src/react-query/TethysApiProvider/TethysApiProvider.tsx - Provides site config through context
>   4. Finally used in apps/lrauv-dash2/lib/useGoogleMaps.ts:
>   const { siteConfig } = useTethysApiContext()
>   const googleMapsApiKey =
>     process.env.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY ??
>     siteConfig?.appConfig.googleApiKey

I think if anything the main problem was that we were injecting a static key so if this static key was not current and we had that in our local .env that would create a problem. Can we confirm that simply removing the process.env injected var here didn't resolve the issue? And if not why is it necessary to make so many additional modifications to get the key? Are we not able to resolve the issue here at the source with less effort?

jimjeffers:
One more thing - the elevator hook relied on the same context to load the key:

> In the useGoogleElevator hook, the Google Maps API key is not directly loaded. Instead, it relies on the Google Maps API being already loaded through the useGoogleMaps hook:
> 
>   1. In useGoogleMaps.ts (line 5-32), the API key is loaded using:
>   const { siteConfig } = useTethysApiContext()
>   const googleMapsApiKey =
>     process.env.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY ??
>     siteConfig?.appConfig.googleApiKey
>   2. Then useGoogleMaps initializes the Google Maps API with:
>   new Loader({
>     apiKey: googleMapsApiKey,
>     version: 'weekly',
>     libraries: ['elevation'],
>   })
>   3. The useGoogleElevator hook (line 3-47) simply checks if the Google Maps API is available and creates an ElevationService instance:
>   const elevator = useMemo(() => {
>     if (
>       typeof google !== 'undefined' &&
>       google.maps &&
>       google.maps.ElevationService
>     ) {
>       return new google.maps.ElevationService()
>     }
>     return null
>   }, [])
> 
>   So the key loading happens in useGoogleMaps, which useGoogleElevator depends on being called first.


It seems like everything was depending on our original useGoogleMaps hook to obtain the API key. Effectively a single point of failure. Can you review what was problematic about the original approach? Was the context not storing the key? What was the original problem? I'm concerned we're breaking from our design patterns and sprinkling embedded API requests directly in client code now when we've had a separation of concerns setup for this all to be facilitated in the api-client.

ksalamy:
No, not a static key being injected. I removed the GoogleMapApiKey from my .env.local for testing on my local system. I even have a debugging response to give me the first 5 characters of the key via the pipeline with tethys and it is accurate and loading correctly - on my local system. 

So the original approach appeared to be working on my local system BUT when I loaded the working fix to dash5-staging I received an invalidKey response from GoogleMaps, no Google Map and no elevator service.

Most of the changes in this PR are to resolve various errors in the dash5-staging console following the last update with the Working ApiKey (on my local system).  I am not sure why things are not loading on the server, hence a lot of debugging code.  A look at the Network on dash5-staging suggests that the "response from server" is taking a long time to occur. That is most likely because it is not resolving the ApiKey.....

If you have the time, please see if you can get the GoogleMapsApiKey and ElevatorService to load and work. I absolutely do not want to affect the design pattern for the project but this issue is mission critical for the team. I appreciate your thoughts.

jimjeffers:
@ksalamy please pull my branch in #418 and see if that works for you locally. I'm worried you've inadvertently introduced new issues with all of this additional work around code as removing the env var based map key seems to be all that is needed to get google maps working again. This branch is based off of commits further back in history before other changes which were merged to alter the maps behavior.

ksalamy:
Yes, this is building fine on my local system.

Karen

_______________
Karen A. Salamy
Associate SW Engineer, MBARI
P 831.775.1820,   F 831.775.1620
[https://www.mbari.org/signature/MBARI-Email-Signature-logo.gif]
[https://www.mbari.org/signature/VertBar.jpg]
Monterey Bay Aquarium Research Institute (MBARI)
7700 Sandholdt Road, Moss Landing CA 95039
www.mbari.org<http://www.mbari.org/>
Advancing marine science and engineering to understand our changing ocean.


"If we knew what it was we were doing, it would not be called research, would it?" - Albert Einstein (1879-1955)

P Please consider the environment before printing this e-mail.

________________________________

Issue #416, opened 2025-05-01
Commands cannot be scheduled


ZackLyon:
**Describe the bug**
When scheduling a command on staging, a toast consistently pops up notifying the user that there's been an error scheduling the command.

**To Reproduce**
1. Go to staging
2. Click on the sim vehicle tab
3. Go to the interface for scheduling command
4. Try scheduling a command
5. See error toast

**Expected behavior**
The user should be able to schedule commands

ZackLyon:
It looks like this issue is related to whether the commands should be sent via cell or sat. Currently, dash5 isn't specifying which method should be used and the endpoint is giving back a 500 status without the via parameter.

ZackLyon:
Closed via #423 (and fixes on the backend that made the via field optional again)

PR #423, opened 2025-05-05
Add communication type and cell timeout controls for scheduling commands

### LINKED ISSUE #422: Add communication type options when scheduling commands
When scheduling commands and missions, there are no options for how to communicate with the vehicle (ie cell, cell then sat, or just sat).

Add controls to allow the user to decide how to communicate with the vehicle when scheduling commands, along with a timeout value for cell communications.

### Dash4 controls for sending via cell or sat
Image

ZackLyon:
- Add controls to select between cell, cell then sat, or just sat to communicate with vehicles
- Add timeout field to select how many minutes to wait while attempting cell communications
- Rename MissionModal and CommandModal UI components to MissionModalView and CommandModalView to avoid confusion with web components with the same names
- Create useScheduleContext to avoid excessive prop drilling of state related to scheduling
- Directly trigger success and failure toasts when sending a command by using the native onSuccess and onFailure methods instead of a useEffect

Closes #422 and #416 

Note: I prioritized this feature since command scheduling was not working due to the `via` field (communication type) becoming required for a bit due to an issue on the backend. I alerted Carlos about it since I was unsure if it was intentional and he quickly resolved it.  I was already in the middle of making this needed update though.


heavyB (approved):
Looks good to me at first glance. Let's take it for a spin!

jimjeffers:
@ZackLyon agreed with @heavyB - it looks good to me too. As long as the renamed components aren't breaking anything elsewhere (I think typescript would have caught that) you should be good to merge.

jimjeffers (approved):
Looks good to me @ZackLyon

Issue #313, opened 2024-02-16
Log history truncated


sqshemet:
**Describe the bug**
I only seem to be able to see a few hours/limited number of log events, even when clicking the "historical logs" button. 

**To Reproduce**
1. Go to "Logs" in any deployment
2. Scroll down
3. See that log history ends well before the beginning of the deployment

**Expected behavior**
See log events at least up to the beginning of the deployment

**Screenshots**
This is the bottom of the log pane, the deployment started about 36 hours earlier
![image](https://github.com/mbari-org/dash5/assets/1900768/7112931f-af3f-4fef-82f7-02b18603888a)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Ubuntu 22.04
 - Browser [e.g. chrome, safari] Firefox
 - Version [e.g. 22] 122.0

jimjeffers:
Researching this issue - the API looks like it only goes back until the 23rd even though we're requesting logs back stopped until the 18th:

```
https://okeanids.mbari.org/TethysDash/api/events?vehicles=aku&eventTypes=command%2Crun&from=2025-02-18T10%3A31%3A15.301-08%3A00&to=2025-02-26T10%3A27%3A46.007-08%3A00
```

Here is the response:

```
{
    "result": [
        {
            "eventId": 23979375,
            "vehicleName": "aku",
            "unixTime": 1740594455054,
            "isoTime": "2025-02-26T18:27:35.054Z",
            "eventType": "command",
            "user": "Quinn Shemet",
            "note": "Vehicle recovered",
            "data": "restart logs",
            "text": "restart logs"
        },
        {
            "eventId": 23979345,
            "vehicleName": "aku",
            "unixTime": 1740592939183,
            "isoTime": "2025-02-26T18:02:19.183Z",
            "eventType": "command",
            "user": "Quinn Shemet",
            "data": "stop",
            "text": "stop"
        },
        {
            "eventId": 23979261,
            "vehicleName": "aku",
            "unixTime": 1740591855300,
            "isoTime": "2025-02-26T17:44:15.3Z",
            "eventType": "run",
            "user": "Quinn Shemet",
            "note": "1k for recovery",
[...]

PR #428, opened 2025-05-09
Paginate fetching logs for Logs and Comms sections of deployment page

### LINKED ISSUE #343: Show data outside of deployment
While vehicles are always on a deployment at sea, they're normally not while in the lab, and we still need to be able to see logs/commands/data for vehicle preparation, testing, etc. Dash5 seems to only show the most recent deployment (or previous deployments if selected). This is the landing page for Daphne, which has been running missions in the lab today in preparation for deployment:

Image

Image

Some way to show all data received from the vehicle, even if not on a deployment. Probably like a "recent' view or something? It would be good if the landing page for any currently-running vehicle showed its most recent data by default, whether that's on a deployment or not. If a vehicle is not currently running I think it's probably okay/good to land on the most recent deployment, instead of arbitrary stale lab data.

**Describe alternatives you've considered**

**Additional context**
I think this is something between "critical" and "necessary" - probably beta testers can still use the dash for ops without this, but we certainly can't use it in the lab.

ZackLyon:
- Create useInfiniteEvents hook that paginates retrieving more events
- Update LogsSection and Comms section components to use the new events hook to retrieve either all logs or just deployment logs
- Create LoadMore button for use with paginated lists

Closes #343 and #313

Note: Brian mentioned that the Schedule section also only goes back to the beginning of the last deployment, but it's a separate issue. I'll open another GH issue for it.


apps/lrauv-dash2/components/CommsSection.tsx
```diff
@@ -23,42 +22,83 @@ export interface CommsSectionProps {
...
+    data,
+    isLoading,
+    isFetching,
+    isFetchingNextPage,
+    fetchNextPage,
+    hasNextPage,
+    refetch,
+  } = allLogs ? allLogsResponse : deploymentResponse
+
+  const flatData = useMemo(() => data?.pages.flat() ?? [], [data])
```
jimjeffers: This could get expensive if we start loading a lot of events. And this will recalculate anytime data updates. You might want to try memoizing on pages or one of these approaches:

 1. Use a more granular memoization that only recalculates when pages change:
 
 ```
  const flatData = useMemo(() => {
    if (!data?.pages) return []
    // Only flatten when the pages array reference changes
    return data.pages.flat()
  }, [data?.pages])
```

  2. Consider windowed rendering or progressive loading:
  
  ```
  // Only flatten visible pages, not the entire dataset
  const visibleData = useMemo(() => {
    if (!data?.pages) return []
    return data.pages.slice(0, visiblePageCount).flat()
  }, [data?.pages, visiblePageCount])
```

  3. Avoid unnecessary re-renders by using a cache key based on page count:
 
 ```
  const flatData = useMemo(() => data?.pages.flat() ?? [],
    [data?.pages?.length, data?.pages])
```

The idea is to get one of these approaches working such that the array only get's built when page data actually updates vs data in general. But you'd have to try it out and see if it's too strict and we get stuck with unwanted cached data too easily. If so it can stay as it is.

File: packages/api-client/src/react-query/Event/useInfiniteEvents.ts [hunk 2]
```diff
@@ -0,0 +1,38 @@
+import { useInfiniteQuery } from 'react-query'
+import { getEvents, GetEventsParams } from '../../axios'
+import { useTethysApiContext } from '../TethysApiProvider'
+
+const DEFAULT_LIMIT = 500
+
+export const useInfiniteEvents = (params: GetEventsParams) => {
```
jimjeffers, line 7: Adding a simple unit test (could be generated via AI) for this hook would be beneficial.

jimjeffers (approved):
Just some minor comments for your consideration. Otherwise nothing actually blocking this from being merged.

apps/lrauv-dash2/components/CommsSection.tsx
ZackLyon: Good call! I updated it using the first strategy and it works well.

ZackLyon:
I'm going to merge this since I addressed the issues Jim had mentioned.

Issue #430, opened 2025-05-20
Display method of communication in Comms Section


ZackLyon:
**Is your feature request related to a problem? Please describe.**
When a command is sent to a vehicle, it can be sent by cell, satellite or an attempt by cell followed by satellite. For communications that involve cell, a timeout is also defined that limits how long the attempt to contact the vehicle by cell will last. Currently, the commands in the comms section do not display which method is being used or how long the timeout is.

**Describe the solution you'd like**
Display the method of communication and if cell related, also the timeout in minutes (which is how it is defined and used in dash4).

**Describe alternatives you've considered**
Perhaps we may want to use an icon as an indicator in future versions.

ZackLyon:
Closed via #431

PR #431, opened 2025-05-20
Update comms section to provide accurate ACK times and display the method of communication and cell timeout

### LINKED ISSUE #272: [Comms] - Vehicle ACK shows incorrect time
**Describe the bug**
Deployment Daphne 03/15/2023:
Monique -
Vehicle ACK does not work for satellite, cf screenshots:

In reality, calibrate_sparton_compass was received via cell and the mission started even before daphne received the mbts_sci2 via cell (see below). Real Ack by daphne for mbts_sci2 was 17:12.

**Screenshots**

CommsQueue2

CommsQueue1 5

CommsQueue

ZackLyon:
Update comms section to provide accurate ACK times and display the method of communication and cell timeout if relevant

- Create useCommsEvents hook to correctly verify status of command/mission communications (queued / sent / ACK / timeout reached), as well as determine what method of communication was used and timeout if via cell. 
- Update the timestamp associated to commands to reflect the timestamp associated to the current status (ie if it's queued, it'll be the command timestamp and if it's acknowledged, it will be the ACK timestamp). This should resolve the mismatch between ACK times between dash4 and dash5.
- Update CommsCell to display communication method and cell timeout, as well as move some of the presentational code from CommSection to CommsCell.
- Update CommsCell to display timeout icon when cell timeout expires
- Refactor CommsSection to use the useCommsEvents hook.
- Refactor the toggle for switching from deployment logs to all logs into its own LogsToolbar component for use across multiple of the sections. (This was meant to be part of the update to add pagination to the Schedule Section)

Additional Notes: I've followed as closely as I could how dash4 determines the ACK state. For cell, it considers the ACK to be when an associated sbdSend is created for the command. For sat, it's when the sbdReceive is confirmed.

Closes #272 and #430 

Screenshot 2025-05-22 at 5 04 21 PM


ZackLyon:
This is **not ready** for merge yet. I still need to make a fix to how I'm determining sat ACK.

ZackLyon:
@jimjeffers - Heads up: this is still pending review when you have time.

apps/lrauv-dash2/lib/useCommsEvents.ts
```diff
@@ -0,0 +1,198 @@
...
+const digitsForIdRegEx = /\d+/
+const timeoutRegEx = /timeout:(\d+)min/
+const timeoutExpiredRegEx = /id=(\d+):\s*Timeout while waiting/
+
+const viaRegEx = /via:\s*(cellsat|cell|sat)(?:,|\])/
+
+const getVia = (note?: string) =>
+  note?.match(viaRegEx)?.[1] as 'cellsat' | 'cell' | 'sat' | undefined
+
+export const useCommsEvents = ({
```
jimjeffers: Have you considered refactoring some of the logic in this hook to helper methods? I think it could be useful to extract some of the logic and add some unit tests for each to ensure we're classifying things reliably.

packages/react-ui/src/Cells/CommsCell.tsx
```diff
@@ -2,30 +2,83 @@ import React from 'react'
...
+import { faStopwatch } from '@fortawesome/free-solid-svg-icons'
 import { IconProp } from '@fortawesome/fontawesome-svg-core'
 import { swallow, truncate } from '@mbari/utils'
 import { AcknowledgeIcon } from '../Icons/AcknowledgeIcon'
+import { ConnectedIcon } from '../Icons/ConnectedIcon'
+import { CommandType } from '../types'
+
+const getStatusTexts = (
+  status: 'queued' | 'sent' | 'ack' | 'timeout',
+  vehicleName?: string,
```
jimjeffers, line 14: Since all of these are potentially null we should have a default value or log them in the template strings below differently so the logs don't state something like `Sent via undefined` etc.. It looks like this could happen for `vehicleName`, `via`, and `timeout`.

apps/lrauv-dash2/lib/useCommsEvents.ts
```diff
@@ -0,0 +1,198 @@
...
+    refetch,
+  } = useInfiniteEvents(params)
+
+  const flatData = useMemo(() => {
+    if (!data?.pages) return []
+    return data.pages.flat()
+  }, [data?.pages])
+
+  const { commands, sbdSendMap, sbdReceiptMap, sbdReceiveMap, timeoutMap } =
+    useMemo(() => {
```
jimjeffers: Claude pointed out this tidbit which I find worth considering:

>  1. Complex memoization in useCommsEvents
>   (apps/lrauv-dash2/lib/useCommsEvents.ts:67-100):
>   const { commands, sbdSendMap, sbdReceiptMap, sbdReceiveMap, timeoutMap } = useMemo(() =>
>    {
>     // Heavy computation creating multiple maps
>   }, [flatData])
>     - Creating 5 different maps on every flatData change could be expensive
>     - Consider breaking this into smaller, more targeted memos
>   2. Nested memoization (apps/lrauv-dash2/lib/useCommsEvents.ts:103-186):
>     - The updatedCommands memo depends on all the maps from the previous memo
>     - This creates a cascade effect where any change triggers full recomputation

```
1. Performance Optimization

  // Consider splitting the heavy memo:
  const commandsOnly = useMemo(() =>
    flatData.filter(e => e.eventType === 'command' || e.eventType === 'run'),
    [flatData]
  )

  const sbdEvents = useMemo(() =>
    flatData.filter(e => e.eventType.startsWith('sbd')),
    [flatData]
  )

  2. Extract Status Logic

  // Create a separate utility file for status determination
  const determineCommandStatus = (command, sbdSendMap, sbdReceiptMap, sbdReceiveMap, 
  timeoutMap) => {
    // Move complex status logic here
  }
```

jimjeffers (changes requested):
Take a look at the suggestions inline -- at the very least I would consider adding tests for the useCommsEvents hook to ensure it's working as you expect since this is going to be a pretty key piece of functionality moving forward.

apps/lrauv-dash2/lib/useCommsEvents.ts
ZackLyon: I extracted the status determination into its own file. On further consideration of the memoized maps, I believe it's better to keep them together in a single memoization since they are all triggered by the same data dependency. Splitting them up would involve mapping or filtering through each type instead of doing a single pass across the data. Let me know if you have a better idea about how to make this more performant.

ZackLyon: I also added tests for the util function that determines command status.

jimjeffers (approved):
These changes look good to me @ZackLyon - merge when ready!

Issue #287, opened 2023-04-18
Can't copy-paste text from log


sqshemet:
**Describe the bug**
Nothing happens when I click and drag to highlight log text to copy it. This is really important for e.g. asking questions in Slack

**To Reproduce**
1. Go to Log (either in sidebar or fullscreened)
2. Click anywhere and hold
3. Move mouse to anywhere
4. Nothing happens :(

**Expected behavior**
Log text should be highlighted and copyable

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser: Chrome
 - Version 109.0.5414.119 (Official Build) (64-bit)

ZackLyon:
@ksalamy - Is this closed by #435 ?

sqshemet:
It seems like the current behavior is that clicking on a single log event copies it to the clipboard? e.g.
```Critical
Time: 11:58
Date: Today
Direction: Download

[CBIT] Communications Fault in component: DockingStepper
```

I think we at least need to be able to copy multiple log events at once. In my opinion it'd be nice to be able to select text "normally" and therefore be able to copy partial log events, (e.g. "Communications fault in component: DockingStepper") but I don't think it's strictly necessary. I'd also vote for not including additional headers and not making the direction copyable. I'd naively expect the example event to copy like "Critical 11:58 Today [CBIT] Communications Fault in component: DockingStepper"

Personally I'd *want* it to copy something like "2025-10-02 11:58:ss Critical [CBIT] Communications Fault in component: DockingStepper", but I might be missing earlier discussion on how to format the new log (@heavyB ?)

heavyB:
Agreed we often copy multiple events when responding to questions with PICs. It seems cumbersome to have to concat

jimjeffers:
Do you want the ability to explicitly select certain logs or just a button to download a txt file of all of the logs currently loaded into the component?

sqshemet:
I think we need the ability to explicitly select certain logs. IMO it's fine if that's limited to adjacent logs within the current display, I don't think we need arbitrary selection. Just something that is as capable as a "normal" copy-paste.

ZackLyon:
Closed via #504

PR #434, opened 2025-05-20
Fix for issue Can't copy-paste text from log #287

ksalamy:
Updated CellVirtualizer.tsx to allow copy and paste of the vehicle log"virtualized-list" (Issue #287). Testing shows it working on my system.

### DIFF
--- a/apps/lrauv-dash2/components/VehicleList.tsx
+++ b/apps/lrauv-dash2/components/VehicleList.tsx
@@ -65,6 +65,7 @@ const ConnectedVehicleCell: React.FC<{
     },
     {
       enabled: !!lastDeployment?.lastEvent,
+      staleTime: 10000, // 10 seconds
     }
   )
   const { data: vehicleInfo, isLoading: vehicleInfoLoading } = useVehicleInfo(
@@ -84,6 +85,10 @@ const ConnectedVehicleCell: React.FC<{
   )
   const pingEvent = useTethysSubscriptionEvent('VehiclePingResult', name)
 
+  // Stabilize vehicle Widget intermittent flashing
+  const dataIsLoading = positionLoading || vehicleInfoLoading
+  const [stableView, setStableView] = useState(false)
+
   // TODO: Remove this demonstations of 'usePlatforms'
   // const { data: platforms } = usePlatforms(
   // { refresh: 'y' },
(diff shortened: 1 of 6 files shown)


jimjeffers (changes requested):
@ksalamy I don't think this is the right approach for a few reasons - primarily we're changing a lot of things about the core component that could be useful or could be problematic without a lot of testing. We're also introducing custom css when we want to be relying on tailwind or simple `style={}` for simple maintainability. 

The good news is that it looks like the problem is actually very simple! The approach here is trying to work around a core browser behavior. You should try a new branch and simply adjust the cell such that the content is wrapped in a div or list element instead of a `button` and the text will be selectable.

CleanShot 2025-05-21 at 09 03 32@2x

I would recommend closing this PR and trying to approach this issue from the cell component the log list relies on.

jimjeffers:
@ksalamy here is where you can adjust the wrapping element:
https://github.com/mbari-org/dash5/blob/develop/packages/react-ui/src/Cells/LogCell.tsx#L36

You could either:

1. Add a 'selectable' prop and use that as a boolean flag to swap between a button or standard div element.
2. Just adjust the onClick property of the [LogsSection to copy the contents of the cell on click](https://github.com/mbari-org/dash5/blob/aa1e0d3cf970b3fef7bd1b9509cf902b5d8d4fe4/apps/lrauv-dash2/components/LogsSection.tsx#L126-L133).

```tsx
import { toast } from 'react-hot-toast'
....
<LogCell
  className="border-b border-slate-200"
  date={date}
  time={time}
  label={displayNameForEventType(item)}
  log={formatEvent(item, siteConfig?.appConfig.external.tethysdash ?? "")}
  isUpload={isUploadEvent(item)}
  onClick={() =>
    navigator.clipboard.writeText(
      formatEvent(item, siteConfig?.appConfig.external.tethysdash ?? "")
    )
    toast.success('Log copied to clipboard.')
  }
/>
```

ksalamy:
Hi Jim,

I have added an update to issue #287 and have sent it to you for review.

Karen

_______________
Karen A. Salamy
Associate SW Engineer, MBARI
P 831.775.1820,   F 831.775.1620
[https://www.mbari.org/signature/MBARI-Email-Signature-logo.gif]
[https://www.mbari.org/signature/VertBar.jpg]
Monterey Bay Aquarium Research Institute (MBARI)
7700 Sandholdt Road, Moss Landing CA 95039
www.mbari.org<http://www.mbari.org/>
Advancing marine science and engineering to understand our changing ocean.


"If we knew what it was we were doing, it would not be called research, would it?" - Albert Einstein (1879-1955)

P Please consider the environment before printing this e-mail.

________________________________

jimjeffers:
Closing this in favor of #435

PR #435, opened 2025-05-21
UPDATE: Fix for issue Can't copy-paste text from log #287. 

ksalamy:
Edited LogsSection to allow a copy from clipboard for each cell in the vehicle logs with an onSelect().  This is based upon suggestion by J. Jeffers. I have tested this and it is working on my system.

### DIFF
--- a/apps/lrauv-dash2/components/LogsSection.tsx
+++ b/apps/lrauv-dash2/components/LogsSection.tsx
@@ -11,6 +11,7 @@ import {
   SubIcon,
   LoadMoreButton,
 } from '@mbari/react-ui'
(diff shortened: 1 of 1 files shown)


jimjeffers (approved):
Excellent! I do have one nitpick if you have time -- just move the block you have in onSelect above the return () and pass it in as a variable. This is just best practice in case we need to memoize the handler down the road.

ksalamy:
Hi Jim,  Sorry just seeing this comment. Will fix and update.

ksalamy:
Hi Jim, I have updated LogsSection with your requested change in code handling. Please review and I will merge it into upstream.

PR #443, opened 2025-06-10
Fix command scheduling issues

ZackLyon:
Resolved the following issues with scheduling commands:

- There are now mission name, component and element options for commands that use them.
- Commands that are scheduled should quickly appear in the scheduling log.
- Fields previously marked token, int or string will now use the appropriate label to describe what they represent.
- Quoted string fields now resolve to quoted strings in the command string.
- Command strings are now assembled without extra characters. 

#### Correctly assembles command text without extra characters
Screenshot 2025-06-09 at 5 40 53 PM

#### Quoted strings are quoted in command text
Screenshot 2025-06-09 at 5 40 04 PM

#### Component and element options are loaded correctly
Screenshot 2025-06-09 at 5 39 05 PM

#### Generic fields have appropriate labels to describe what they represent
Screenshot 2025-06-09 at 5 37 58 PM

#### Mission options are loaded correctly
Screenshot 2025-06-09 at 5 53 31 PM

### DIFF
--- a/packages/react-ui/src/Modals/CommandModalSteps/BuildTemplatedCommandStep.tsx
+++ b/packages/react-ui/src/Modals/CommandModalSteps/BuildTemplatedCommandStep.tsx
@@ -52,7 +52,7 @@ const argumentAsParameter = (
     case 'ARG_INT':
       return {
         argType: arg.argType,
-        name: 'int',
+        name: arg?.altName ?? 'int',
         description: 'An integer value',
         inputType: 'number',
         value,
@@ -78,7 +78,7 @@ const argumentAsParameter = (
     case 'ARG_STRING':
       return {
         argType: arg.argType,
-        name: 'string',
+        name: arg?.altName ?? 'string',
         description: 'A string value',
         inputType: 'string',
         value,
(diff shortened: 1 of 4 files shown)


packages/react-ui/src/Modals/CommandModalSteps/BuildTemplatedCommandStep.tsx
```diff
@@ -331,10 +332,16 @@ export const BuildTemplatedCommandStep: React.FC<
               ?.slice(-2)
               .join('.')
           : undefined
+      case 'ARG_UNIVERSAL':
+        return values?.slice(-1)[0] ?? undefined
       case 'ARG_COMPONENT':
         return value ? values[1] : value
```
jimjeffers, line 338: This isn't you change but it's noticeable when reviewing the code that this could potentially crash as we are only checking for the presence of value not values and more importantly we're not testing to ensure values contains two elements! Yikes!

jimjeffers (approved):
This looks great Zack. I don't see anything that needs to be changed. Some of the existing code -- not your changes could potentially be updated but that's not really in the scope of this PR.

packages/react-ui/src/Modals/CommandModalSteps/BuildTemplatedCommandStep.tsx
jimjeffers: This is referring to line 338 `return value ? values[1] : value`

PR #444, opened 2025-06-13
Fix missing or partially displayed mission names

ZackLyon:
- Fix regex to extract mission names in useRecentRuns
- Add simple loading and no results states to Table and by extension, MissionTable
- Display truncated mission command string for mission description instead of repeating mission name; hover to view full mission command string
- In MissionModal, useRecentCommands was waiting for useVehicleNames to load all the vehicle names before fetching, but supplying an empty array to useRecentCommands yields the same result without waiting; fixed

### Staging with missing or partial mission names
Screenshot 2025-06-12 at 6 04 18 PM

### Mission names made whole
Screenshot 2025-06-12 at 6 05 18 PM

### DIFF
--- a/apps/lrauv-dash2/components/MissionModal.tsx
+++ b/apps/lrauv-dash2/components/MissionModal.tsx
@@ -146,13 +146,10 @@ const MissionModal: React.FC<MissionModalProps> = ({
     { enabled: !!vehicleName }
   )
 
-  const { data: recentRunsData } = useRecentRuns(
-    {
-      vehicles: vehicles ?? [],
-      from: LAST_60_DAYS,
-    },
-    { enabled: !!vehicles }
-  )
+  const { data: recentRunsData, isLoading: recentRunsLoading } = useRecentRuns({
+    vehicles: [], // All vehicles by default
+    from: LAST_60_DAYS,
+  })
   const {
     mutate: createCommand,
     isLoading: sendingCommand,
@@ -207,6 +204,7 @@ const MissionModal: React.FC<MissionModalProps> = ({
       ?.map(
         (
           {
+            data,
             mission,
             vehicleName: vehicle,
             isoTime,
@@ -222,7 +220,7 @@ const MissionModal: React.FC<MissionModalProps> = ({
             id: mission,
             category,
             name,
-            description: mission,
+            description: data,
             vehicle,
             ranOn: capitalize(
               DateTime.fromISO(isoTime).toFormat('MMM. d yyyy')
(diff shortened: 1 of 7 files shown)


apps/lrauv-dash2/components/MissionModal.tsx
```diff
@@ -146,13 +146,10 @@ const MissionModal: React.FC<MissionModalProps> = ({
...
 
-  const { data: recentRunsData } = useRecentRuns(
-    {
-      vehicles: vehicles ?? [],
-      from: LAST_60_DAYS,
-    },
-    { enabled: !!vehicles }
-  )
+  const { data: recentRunsData, isLoading: recentRunsLoading } = useRecentRuns({
+    vehicles: [], // All vehicles by default
```
jimjeffers, line 150: All vehicles by default or all vehicles always? This should be fine if that's intended - but just checking.

packages/api-client/src/react-query/Command/useRecentRuns.ts
```diff
@@ -23,7 +23,7 @@ export const useRecentRuns = (
   )
 
   const formattedData = query.data?.map((event) => {
-    const mission = event.data?.match(/[a-zA-Z_/]+(\.xml|\.tl)/)?.[0] ?? ''
+    const mission = event.data?.match(/[a-zA-Z0-9_/]+(\.xml|\.tl)/)?.[0] ?? ''
```
jimjeffers, line 30: It's not a bad idea to document any of these regex patterns with a comment. Especially since they appear to be the crux of some of these issues:

```
// Regex: /[a-zA-Z0-9_/]+(\.xml|\.tl)/
//   • [a-zA-Z0-9_/]+  → one or more letters, digits, “_”, or “/” (captures a simple path or filename)
//   • (\.xml|\.tl)    → that sequence must end with “.xml” **or** “.tl” (the dot is escaped to mean a literal dot)
// Summary → Pulls out the first path-like token in event.data that ends in one of those extensions,
//           e.g. "mission_42.xml", "missions/alpha/001.tl", "surveys/foo_bar.xml".
```

Also - do we have jest setup in the api client? It would be worth adding a test here.

jimjeffers (approved):
Approved but you could add a little documentation and a unit test. (optional)

apps/lrauv-dash2/components/MissionModal.tsx
ZackLyon: If no vehicles are specified when making the request to the endpoint, it will send back events for all vehicles. The way it was set up previously, we were fetching all the vehicle names from the vehicles endpoint, then using them as a param for this endpoint. This just saves us a step. We might want to consider adding a checkbox to limit this to current vehicle as an added feature though.

jimjeffers:
Looks good to me Zack!

ZackLyon:
Thanks for the quick review update, Jim!

Issue #273, opened 2023-03-29
Final Mission Review Screen does not show Waypoints


ksalamy:
**Describe the bug**
Deployment Daphne 03/15/2023:
Quinn -
The final mission review screen doesn’t show waypoints, though the previous two do. 

Also says DVL off, when that’s not true? See screenshots:

![QuinnJIRA1](https://user-images.githubusercontent.com/12836105/228688496-4dc79b75-9fff-4e06-9d77-e4c9a06db556.png)
![QuinnJIRA2](https://user-images.githubusercontent.com/12836105/228688527-56b9beac-ac7b-44fc-98d7-34066ab2eba9.png)
![QuinnJIRA3](https://user-images.githubusercontent.com/12836105/228688549-aead2e50-bb33-470b-b312-793b4b35e460.png)

ZackLyon:
Closed via #450

Issue #267, opened 2023-03-29
Mission "re-run" not working.


ksalamy:
**Describe the bug**
Deployment Galene 9- 01/26/2023:
Brian -
Mission re-run not working.

sqshemet:
More detail:
1. Go to Missions/Recent Runs
2. Select mission to rerun. In the below screenshots I have selected a transit on Ahi from Jan 29th with a set/overridden waypoint and depth.
3. Click through to "waypoitns" and "parameters" dialogues

![Image](https://github.com/user-attachments/assets/6a1aba89-c30a-49b6-9b85-cb69ec937ea4)

Expected behavior:
Overridden waypoints and parameters from previous run of missions are present for reuse or edit
Current behavior: 
All overridden waypoints/parameters are reset

![Image](https://github.com/user-attachments/assets/83261662-ebe2-4953-a74d-4b8c37cbbd46)

ZackLyon:
Closed via #450

PR #450, opened 2025-07-31
Fix mission scheduling

ZackLyon:
Fix issues with mission scheduling

- Rerun-mission flow – now applies parameter & waypoint overrides and preserves/abbreviates units; shows visual cue when a value’s unit was overridden.
- New unit-handling utilities: abbreviateUnits, getUnitFromAbbreviation, and updates to makeValueUnitString; used across API client, utils, and UI.
- API-client enhancement: new /preview request (“usePreview” hook) plus supporting getPreview endpoint and React-Query hook.
- Mission & command text generation now outputs abbreviated units; related logic moved to makeCommand and scheduling components.
- Can pick mission at index 0, reset params when backing out, faster param text entry, boolean parameters rendered human-readably.
- Parameter/mission step refactor; table rows keyed correctly; long names/values wrap; checkbox replaced with dropdown for booleans.
- Select component can accept custom styles, enabling richer styling in mission tables.
- Numerous bug fixes: crash on window refocus, duplicate mission/category state, reserved-param filtering (correctly identify safety and comms params and keep separate), insert handling, etc.
- Added/updated Jest tests for overrides extraction, recent runs, unit utils, parameter tables, and mission modal view.

## Override values and units are now applied when rerunning missions
Screenshot 2025-07-31 at 3 10 41 PM

## Insert and root level commands with the same command name are managed separately (in this example, StandardEnvelopes.MinAltitude and MinAltitude are updated separately). Also, waypoints now appear in the final summary review screen.
Screenshot 2025-07-31 at 3 12 03 PM

Screenshot 2025-07-31 at 3 12 22 PM


ZackLyon:
@jimjeffers - Please don't murder me! I promise to break things down into smaller chunks going forward. Mission scheduling is in a much better place now, though. I'm not an expert, so I may be missing some edge cases or nuances that I don't understand, but this creates missions very similarly to dash4 now.

Notes:
- I'm not sure if the parameters are in the correct order in the scheduling. I didn't change anything with that, so it's the same as it was.
- When you are rerunning a mission and a parameter has an override, if the user changes the override back to the default value, it still registers as an override and will be included in the command text. I can't imagine it matters other than housekeeping, but it is a loose thread. Once again, this is the same as before.
- It's probably not a priority, but parameter fields should be locked down so users can't be inappropriately using negative numbers, random characters, etc

ZackLyon:
References #267 and #273

jimjeffers:
Hey @ZackLyon I've added two issues to follow up on this PR as stacked PRs:

https://github.com/mbari-org/dash5/issues/455
https://github.com/mbari-org/dash5/issues/456

This should address my concerns regarding the overall implementation. Otherwise we'll just have to do some QA after this bad boy is merged to see if any other subsequent issues need to be opened up. I'm sure this will be a huge improvement over what is currently live.

jimjeffers:
I traced the mission-name sort regression to `15f4df5` in `packages/react-ui/src/Modals/MissionModalSteps/MissionStep.tsx` after sorting moved into `useMemo`:

```ts
if (!sortColumn) return filteredMissions
```

`MISSION NAME` is sort column `0`, so clicking that header sets `sortColumn = 0`, which this condition treats as "no sort". The minimal fix is:

```ts
if (sortColumn == null) return filteredMissions
```

That still preserves `null`/`undefined` as the unsorted state, but allows `0` to be treated as a valid first-column index so the sort actually runs.

jimjeffers:
There is also a stale-override issue in `packages/react-ui/src/Modals/MissionModalSteps/hooks/useManagedParameters.ts`: `overrideMap` is initialized from `defaultOverrides`, but the sync effect only re-applies them when the map is empty.

```ts
// User picks Recent Run A
defaultOverrides = {
  MissionTimeout: { value: '4' },
  NeedCommsTime: { value: '60' },
}
overrideMap = buildOverrideMap(defaultOverrides)

// User then picks Recent Run B in the same modal
defaultOverrides = {
  MissionTimeout: { value: '8' },
  NeedCommsTime: { value: '90' },
}

// Effect does nothing because overrideMap is not empty
if (Object.keys(overrideMap).length === 0) {
  setOverrideMap(buildOverrideMap(defaultOverrides))
}
```

So `updatedParameters` can still render Run A's values:

```ts
updatedParameters = applyOverrides(parameters, overrideMap)
// MissionTimeout stays 4 instead of 8
// NeedCommsTime stays 60 instead of 90
```

The fix is to rebuild/reset `overrideMap` whenever the selected run or `defaultOverrides` changes. That works because the derived parameter lists already read entirely from `overrideMap`, so replacing that map cleanly swaps the UI to the newly selected run.

jimjeffers:
There is also a selection-state issue in the Mission step: the modal can open with a preselected mission from `globalModalId.meta.mission`, but `selectedMissionCategory` is initialized to `Recent Runs`, and `MissionStep` immediately clears `defaultSearchText` on mount. That leaves a state where the selected mission is not visible in the filtered table, but `selectedId` is still truthy so `Next` remains enabled.

The fix is to derive the initial category/search state from the preselected mission, or clear `selectedId` whenever the selected category/filter no longer contains it. That works because it keeps the visible table state and the wizard's selection state in sync, so the user cannot advance with a hidden selection.

Issue #447, opened 2025-06-27
Unable to sign in as PIC


ksalamy:
**Describe the bug**
Brian K - Unable to sign in as PIC on dash5 for vehicle Galene

**To Reproduce**
1. Log in as Dash5 user
2. Go to individual vehicle page
3. Click on "No PIC / No On-Call" button in upper right header area
4. The popup modal shows no option to sign up for PIC or On-Call PIC

**Expected behavior**
A join link should show for both PIC and On-Call which the user can then add themselves to or view other users currently listed as PIC or On-Call.

**Screenshots**
![Image](https://github.com/user-attachments/assets/42753e5a-a248-4690-b750-3d49cbb68a1c)

ZackLyon:
Closed via #452

PR #452, opened 2025-08-04
Verify all tracked vehicles are included when assigning PIC and OnCall

ZackLyon:
- Fix bug where only vehicles with recent sign in or out events were being displayed in the PIC/OnCall role assignment modal


ZackLyon:
Closes #447

packages/api-client/src/react-query/User/useVehiclePicAndOnCall.ts
```diff
@@ -45,7 +45,10 @@ export const useVehiclePicAndOnCall = ({
   vehicleName,
   enabled = true,
 }: UseVehiclePicAndOnCallParams): UseVehiclePicAndOnCallResult => {
-  const vehicleNames = Array.isArray(vehicleName) ? vehicleName : [vehicleName]
+  const vehicleNames = useMemo(
```
jimjeffers, line 50: Good idea.

packages/api-client/src/react-query/User/useVehiclePicAndOnCall.ts
```diff
@@ -84,17 +87,21 @@ export const useVehiclePicAndOnCall = ({
...
-    if (!query.data?.length) return undefined
+    // Ensure every requested vehicle is represented, even if no sign in or out events exist
+    const eventsByVehicle: Record<string, GetEventsResponse[]> = {}
+
+    vehicleNames.forEach((name) => {
+      eventsByVehicle[name] = []
+    })
 
-    // Group events by vehicle
-    const eventsByVehicle = query.data.reduce((acc, event) => {
```
jimjeffers, line 92: Was the issue with the reduce fn simply that events for vehicle names without any events were not getting populated as blank arrays? Just curious. Looks good Zack.

jimjeffers (approved):
Ship it!

packages/api-client/src/react-query/User/useVehiclePicAndOnCall.ts
ZackLyon: Yes, that was the issue if I remember correctly. The reducer was early returning if there were no events for a specific vehicle.r

Issue #446, opened 2025-06-27
Filtering a vehicle's logs by "SatComms" displays direct "CellComms" too.


ksalamy:
**Describe the bug**
Brian K - When filtering the vehicle logs for Sat Comms, Cell Comms logs also appear. Only Sat Comms should be displayed.

**To Reproduce**
On an Individual vehicle deployment page, 
  1. Click on "Log" 
  2. In the dropdown "Filter by event type" area - choose "Sat Comms".

**Expected behavior**
When filtering the vehicle logs for Sat Comms, only Sat Comms events should be display not Cell Comms (Direct Comms).

**Screenshots**
![Image](https://github.com/user-attachments/assets/140ca31b-ff79-4eb6-8038-6eac3fec4aad)

ZackLyon:
Closed via #453

PR #453, opened 2025-08-04
Apply comms filtering in Logs section

ZackLyon:
The Logs section filters were only filtering by event type, however comms events share event types such as `sbdSend` and `sbdReceive` and therefore, need additional filtering

- Created util functions to filter by event type with the additional comms filters included (this is set up so that if more filters are included in the future, it will still apply them correctly)
- Updated LogsSection with the additional filters

### Sat Comms
Screenshot 2025-08-04 at 4 24 57 PM

### Direct Comms
Screenshot 2025-08-04 at 4 25 26 PM

### DIFF
--- a/apps/lrauv-dash2/components/LogsSection.tsx
+++ b/apps/lrauv-dash2/components/LogsSection.tsx
@@ -17,6 +17,7 @@ import formatEvent, {
   eventFilters,
   isUploadEvent,
 } from '../lib/formatEvent'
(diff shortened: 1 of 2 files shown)


ZackLyon:
Closes #446

jimjeffers (approved):
Seems fine. I don't think I'll push for unit tests on the helper utils here. Looks pretty straight forward. If we find out there's a bug in these simple helpers later I'll say "Where were the tests!?".

Issue #441, opened 2025-06-05
Improve messaging for PIC and On Call when user is not logged in


jimjeffers:
## Summary
When a user is not logged in, the interface currently displays "no PIC and no On Call" which could be confusing. Instead, it should display "Unavailable" to better indicate that the information is not accessible due to authentication status.

## Current Behavior
- Displays "no PIC and no On Call" for unauthenticated users

## Desired Behavior
- Display "Unavailable" instead of "no PIC and no On Call" when user is not logged in

## User Story
As an operator, I want clearer messaging when PIC and On Call information is not available due to authentication status, so that I understand why the information is not displayed.

## Acceptance Criteria
- [ ] Replace "no PIC and no On Call" message with "Unavailable" for unauthenticated users
- [ ] Ensure authenticated users continue to see actual PIC and On Call information when available
- [ ] Maintain existing behavior for authenticated users when no PIC/On Call is assigned

🤖 Generated with [Claude Code](https://claude.ai/code)

ZackLyon:
Closed via #454

PR #454, opened 2025-08-05
Show Unavailable For PIC / OnCall Roles When User Is Not Logged In

ZackLyon:
- Update createRoleLabel util function and RoleReassignButton with unauthenticated and loading states
- Display Unavailable message for role reassignment button when user is not logged in
- Do not display any role information in Handoff section header when user is not logged in
- Update useVehiclePicAndOnCall hook to accept query param options in the same way as other hooks
- Only enable useVehiclePicAndOnCall on the deployment page when user is authenticated
- Fix issue where empty strings were causing the roles to render incorrectly in the handoff section header (ie instead of reading 'Unassigned/Brian K' when Brian was the OnCall and no-one was PIC, it would read '/BrianK')

### DIFF
--- a/packages/utils/src/createRoleLabel.ts
+++ b/packages/utils/src/createRoleLabel.ts
@@ -1,5 +1,13 @@
 import { shortenName } from './shortenName'
 
+export interface CreateRoleLabelParams {
+  operators: string[]
+  role: 'PIC' | 'On-Call'
+  currentUser?: string
+  authenticated?: boolean
+  loading?: boolean
+}
+
 /**
  * Creates a human-readable label for role assignments based on a list of operators and the current user's status.
  *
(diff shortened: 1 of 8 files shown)


ZackLyon:
Closes #441

packages/utils/src/createRoleLabel.ts
```diff
@@ -8,11 +16,17 @@ import { shortenName } from './shortenName'
...
+export const createRoleLabel = ({
+  operators,
+  role,
+  currentUser,
+  authenticated,
+  loading,
+}: CreateRoleLabelParams) => {
+  if (!authenticated && role === 'PIC') return 'Unavailable'
+  if (!authenticated && role === 'On-Call') return ''
+  if (loading) return '...'
```
jimjeffers, line 28: Nice centralized refactor. This looks good.

jimjeffers (approved):
Nothing concerning here @ZackLyon you are good to merge.

Issue #455, opened 2025-08-05
Refactor MissionModal.tsx: Break down large component into smaller, manageable pieces


jimjeffers:
# MissionModal Refactoring Example

## Current Issues

The `MissionModal.tsx` component is currently 432+ lines and violates the single responsibility principle. It handles:
- Mission selection and categorization
- Parameter override management
- Waypoint calculations
- Distance calculations
- Command scheduling
- State management for multiple data sources

#### `useMissionData.ts`
```typescript
import { useMemo } from 'react'
import { useScript, useRecentRuns, useFrequentRuns } from '@mbari/api-client'
import { getUnitFromAbbreviation } from '@mbari/utils'

export const useMissionData = (selectedMission?: string, selectedCategory?: string) => {
  const { data: selectedMissionData } = useScript({
    path: selectedMission as string,
  }, { enabled: !!selectedMission })

  const { data: recentRunsData } = useRecentRuns()
  const { data: frequentRunsData } = useFrequentRuns()

  const recentRuns = useMemo(() => {
    return recentRunsData?.map(({
      data, mission, vehicleName, isoTime, user, note,
      parameterOverrides, waypointOverrides
    }) => {
      const { category, name } = parseMissionPath(mission)
      return {
        id: mission,
        category,
        name,
        description: data,
        vehicle: vehicleName,
        ranOn: capitalize(DateTime.fromISO(isoTime).toFormat('MMM. d yyyy')),
        ranBy: capitalizeEach(user ?? ''),
        recentRun: true,
        note,
        waypointOverrides,
        parameterOverrides,
        parameterCount: parameterOverrides.length,
        waypointCount: waypointOverrides.length,
      }
    }).filter((mission, index, s) => 
      s.findIndex((m) => m.id === mission.id) === index
    ) ?? []
  }, [recentRunsData])

  return {
    selectedMissionData,
    recentRuns,
    frequentRuns: frequentRunsData, // process similarly
  }
}
```

#### `useParameterOverrides.ts`
```typescript
import { useMemo } from 'react'
import { ParameterProps } from '@mbari/react-ui'
import { getUnitFromAbbreviation } from '@mbari/utils'

export const useParameterOverrides = (
  parameters: ParameterProps[],
  selectedMission?: string,
  selectedCategory?: string,
  recentRuns: any[],
  unitsData?: any[]
) => {
  const parametersWithOverrides = useMemo(() => {
[...]

PR #457, opened 2025-08-08
Refactor MissionModal

ZackLyon:
- Refactor applying parameter overrides, parsing mission text, and determining waypoints each into separate hooks


ZackLyon:
@jimjeffers - I refactored the logic in MissionModal into separate hooks, but left the UI composition as we discussed

ZackLyon:
References #455

jimjeffers (approved):
Looks great Zack! Merge when ready to manually configure a release. I believe the GH action does not successfully create a release by default so you'll need to use the yarn command to generate the version tag.

Issue #456, opened 2025-08-05
Add comprehensive error handling for distance calculations in MissionModal


jimjeffers:
# Distance Calculation Error Handling Solutions

## Current Issues

The distance calculation logic in `MissionModal.tsx` (lines ~105-130) lacks proper error handling and could cause the component to crash or behave unexpectedly when:
- Invalid coordinate data is provided
- Coordinate parsing fails
- Mathematical operations fail
- Turf.js library throws exceptions

## Current Problematic Code

```typescript
// Current implementation in MissionModal.tsx
useEffect(() => {
  const applicableWaypoints = updatedWaypoints.filter(
    (w) =>
      ![
        w.lat?.toLocaleLowerCase() ?? 'nan',
        w.lon?.toLocaleLowerCase() ?? 'nan',
      ].includes('nan')
  )

  if (applicableWaypoints.length > 0) {
    const newDistance = applicableWaypoints.reduce((acc, curr, index) => {
      if (index === 0) return acc

      const prev = applicableWaypoints[index - 1]
      return (
        acc +
        distance(
          point([parseFloat(prev.lon), parseFloat(prev.lat)]),
          point([parseFloat(curr.lon), parseFloat(curr.lat)])
        )
      )
    }, 0)

    if (newDistance !== estDistance) {
      setEstDistance(newDistance)
    }
  } else if (estDistance) {
    setEstDistance(null)
  }
}, [estDistance, updatedWaypoints])
```

## Problems Identified

1. **No error boundaries** around parsing operations
2. **Missing validation** for coordinate ranges
3. **No handling** of NaN results from parseFloat
4. **Potential infinite loops** if Turf.js throws exceptions
5. **No user feedback** when calculations fail
6. **Missing dependency array** in useEffect
7. **Performance issues** - recalculates on every render

### `useWaypointDistanceCalculation.ts`

```typescript
import { useState, useEffect, useCallback } from 'react'
import { point, distance } from '@turf/turf'
import { tryCatch } from '@mbari/utils'

interface WaypointCalculationResult {
  distance: number | null
  error: string | null
  isCalculating: boolean
}

interface WaypointCoordinate {
  lat?: string | number
  lon?: string | number
  [key: string]: any
}

export const useWaypointDistanceCalculation = (
  waypoints: WaypointCoordinate[]
): WaypointCalculationResult => {
  const [result, setResult] = useState<WaypointCalculationResult>({
    distance: null,
    error: null,
    isCalculating: false
  })

[...]

PR #458, opened 2025-08-11
Update useWaypointCalculations and add error messaging

ZackLyon:
- Update useWaypointCalculations to rely on useMemo instead of useEffect since it is clearly tied to a single, limited source of data for its updates
- Add toasts to relay errors that occur to the user if the distance calculation fails


ZackLyon:
References #456

jimjeffers (approved):
This looks fine Zack 👍🏼 Safe to merge!

Issue #460, opened 2025-08-19
Insert mission parameters disappear after inactivity or window blur


ZackLyon:
**Describe the bug**
If the parameter step is left open for too long or if the user mouses away from the browser, the parameters with inserts will disappear and only the root level parameters will remain 

**To Reproduce**
1. Go to mission scheduling
2. Click through to parameter step on a mission with inserts
3. Generally tool around and do anything except fill in the inserts or navigate away from the parameter step

**Expected behavior**
All parameters should be persistent

**Additional context**
The issue likely stems from not adhering to the single source of truth principle. The parameters are being managed in both internal and external state in the parameter step, likely since it initially helped with search filtering and sort order. The issue probably stems from trying to keep these separate states in sync.

ZackLyon:
This issue is still ongoing and was not fixed by #461

ZackLyon:
@carueda - I've been having this issue where the parameters step of mission scheduling sometimes shows the insert parameters and sometimes doesn't. I was able to track the root cause down to the data I'm getting back from the `commands/scripts` endpoint. It looks like that endpoint often gives back the correct information, but occasionally sends back a partial response and in one case, I had it send back additional inserts while still showing as 200 status. I'm using the same parameters every time as you can see from the network logs/json responses below. This seems to happen with relative frequency, unfortunately; maybe one out of 5 times or so. I was able to replicate it once while using the TethysDash also after making repeated requests. Could you please investigate this or advise on how to navigate this? I'm happy to discuss it further and would be glad to answer any questions you have about it.

### Network logs showing differing payload sizes for the same request
<img width="2336" height="161" alt="Image" src="https://github.com/user-attachments/assets/b200f62e-4661-456a-b5bd-6e82b7dab091" />

<img width="2337" height="177" alt="Image" src="https://github.com/user-attachments/assets/f00b3e9f-ba03-4689-a342-00d81372a641" />

### TethysDash response with inserts for sci2
[response_1756755698223.json](https://github.com/user-attachments/files/22098762/response_1756755698223.json)

### TethysDash response without inserts for sci2
[response_1756755594975.json](https://github.com/user-attachments/files/22098731/response_1756755594975.json)

carueda:
@ZackLyon - As we haven't seen anything similar to this in dash4, seems like it is some UI logic in dash5. If you think it is API related, can you try to reproduce with dash4? Also, can you indicate the concrete requests (not the API doc links, but the actual requests with parameters, etc.)?

[...]

Issue #268, opened 2023-03-29
Mission parameter ordering does NOT follow the aggregate structure.


ksalamy:
**Describe the bug**
Deployment Triton 02/16/2023:
Brian - Mission parameter ordering needs to follow the aggregate structure. 

Additionally, the difference between parameters that are inside of an insert and top level mission parameters need to be displayed as such.

jenniferjkim:
A few questions: 

- Besides grouping the parameters, do the parameters need to be ordered with some sort of hierarchy? 
- Can you clarify what you mean by parameters inside of an insert vs top level mission parameters? What are the differences?

heavyB:
Probably best to walk through this via zoom but here's an example from Dash 4 and 5 respectively along with a copy of the mission file (note file extension has been changed to txt but is usually the DSL suffix .tl for TethysL https://docs.mbari.org/tethysl/)

![Image](https://github.com/user-attachments/assets/19b52e36-5ee3-46e3-88c1-08ab25671b49)

![Image](https://github.com/user-attachments/assets/fd7da28c-b1ab-4cdc-8247-c1948dfab92e)

[mbts_sci2.txt](https://github.com/user-attachments/files/18664742/mbts_sci2.txt)

ZackLyon:
@heavyB @jenniferjkim - It looks like this might have gotten (mostly) fixed in my last PR. The parameters are in the correct order, except the waypoints, safety and comms params are all separated out into their own pages, which is a part of the design decision here. Is that all that is needed here? Or should I be inserting the waypoints, safety and comms params back in the same order they initially appeared when the mission text is assembled before scheduling happens?

## mbts_sci2 mission appears in the same order as seen above in dash4 (minus waypoints)
<img width="1017" height="2210" alt="Image" src="https://github.com/user-attachments/assets/d4c5bb5a-a04b-4956-9403-17e48832be94" />

ZackLyon:
Closed via #461

PR #461, opened 2025-08-19
Fix mission parameter ordering and group by insert

ZackLyon:
- Group mission parameters by insert with the root level parameters shown first
- Create AccordionParameterTable component to make parameter groups collapsible
- Add stories and tests for new component
- Resolve issue where parameters that aren't root level would disappear from the Parameter Step screen after a period of inactivity and/or mousing away from the browser for a bit
- Apply parameter override color directly to the input in the ParameterField instead of passing it in from the parent component
- Add a max height option for accordion cell so that they can conform to a shorter height if there's less content while still growing to the max height limit before scrolling

References #268 and #460


ZackLyon:
### Parameters grouped by insert starting with the root parameters:
Screenshot 2025-08-18 at 6 43 50 PM

jimjeffers (approved):
Nice refactor of the parameter step component. Everything looks really solid here. I don't have any concerns from the code and the screenshots look great.

jimjeffers:
@ZackLyon you can merge and release when ready. Looks great!

PR #462, opened 2025-08-22
Updated schedule pane to display last mission with start time

ZackLyon:
- Use mission start time endpoint instead of inferring last mission from command status endpoint
- Display last mission with associated start time in schedule header during active deployments


jimjeffers (approved):
Nice. Looks like we already implemented this hook in the api-client and just didn't use it eh? Looks fine to me Zack.

Issue #463, opened 2025-08-28
Address Mission Parameters  UI Feedback


ZackLyon:
Address the following feedback for the mission parameters UI

- Text overflows root and insert parameter drawers when they are closed (affects Chrome/Firefox; fine in Safari)
- Reduce parameter description font size
- Reduce height of drawers to make it more apparent that the modal is scrollable
- Reduce padding between heading, search box and parameters section
- Add an X cancel button to upper left hand corner of the modal

ZackLyon:
Closed via #464

PR #464, opened 2025-09-02
Mission parameter UI updates

ZackLyon:
Changes requested by the design team for the updated mission parameter step

- Reduce vertical padding between mission text, search bar and mission parameter tables
- Add X button in top right corner of modal as another method to close modal besides cancel button
- Reduce font size of parameter description text
- Reduce height of mission parameter tables to always allow multiple tables to display, so that user knows that more parameter options are available
- Fix text overflowing parameter tables
- Increase modal height for mission scheduling to allow more content to show

### DIFF
--- a/packages/react-ui/src/Tables/AccordionParameterTable.tsx
+++ b/packages/react-ui/src/Tables/AccordionParameterTable.tsx
@@ -82,7 +82,7 @@ export const AccordionParameterTable: React.FC<
                       {name}
                     </span>
                     {description && (
-                      <div className="break-words text-stone-600/60">
+                      <div className="break-words text-sm text-stone-600/60">
                         {description}
                       </div>
                     )}
--- a/packages/react-ui/src/Tables/AccordionParameterTable.tsx
+++ b/packages/react-ui/src/Tables/AccordionParameterTable.tsx
@@ -129,13 +129,14 @@ export const AccordionParameterTable: React.FC<
   return (
     <div
       className={clsx(
-        'flex h-full flex-col border-2 border-stone-200',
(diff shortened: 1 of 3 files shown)


ZackLyon:
References #463

packages/react-ui/src/Tables/AccordionParameterTable.tsx
```diff
@@ -129,24 +129,29 @@ export const AccordionParameterTable: React.FC<
...
-        'flex h-full flex-col border-2 border-stone-200',
+        'flex flex-col border-2 border-stone-200',
+        open && 'h-3/4',
         className
       )}
       style={style}
     >
       <AccordionHeader
-        className="sticky top-0 z-10"
+        className="top-0 z-10"
```
jimjeffers, line 139: Does this break scrolling behavior at all or is that a moot point as there is no scrolling happening in this component? Removing sticky means the element is simply relative - do you still need top-0 and z-10 in that case?

jimjeffers (approved):
Approved - comment on the accordion header and what's happening there after removing position sticky.

packages/react-ui/src/Tables/AccordionParameterTable.tsx
ZackLyon: It doesn't break scrolling inside the accordion section itself. There are accordion sections for the root parameters plus each of the inserts. All those sections scroll, but the modal body also scrolls so you can see all the sections. I removed the sticky so that it wouldn't scroll with the modal body. You're probably right that it doesn't need the top or z index classes anymore.

PR #470, opened 2025-09-02
Mission scheduling parameters disappearing 

ZackLyon:
- Create helper function to limit amount of concurrent api calls
- Limit the amount of concurrent calls to the commands/script endpoint and cache the results in useRecentRuns
- Do not refetch data on window refocus for useRecentRuns or useEvents since they are both computationally intensive
- Memoize the parameters and provide a stale time for useRecentRuns to further minimize refetching in useMissionData hook
- By default, no mission should be selected when scheduling missions

### DIFF
--- a/packages/api-client/src/axios/Util/concurrencyLimiter.ts
+++ b/packages/api-client/src/axios/Util/concurrencyLimiter.ts
@@ -1,0 +1,25 @@
+// Concurrency limiter for script fetches
+export const createLimiter = (concurrency: number) => {
+  let activeCount = 0
+  const queue: Array<() => void> = []
+  const next = () => {
+    if (queue.length === 0 || activeCount >= concurrency) return
+    activeCount++
+    const run = queue.shift()!
+    run()
+  }
+  return <T>(fn: () => Promise<T>): Promise<T> =>
+    new Promise<T>((resolve, reject) => {
+      const task = () => {
+        fn()
+          .then((value) => resolve(value))
+          .catch(reject)
+          .finally(() => {
+            activeCount--
+            next()
+          })
+      }
+      queue.push(task)
+      next()
+    })
+}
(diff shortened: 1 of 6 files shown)


ZackLyon:
References #460

packages/api-client/src/axios/Util/concurrencyLimiter.ts
```diff
@@ -0,0 +1,25 @@
+// Concurrency limiter for script fetches
+export const createLimiter = (concurrency: number) => {
```
jimjeffers, line 2: Fancy. Do we expect to support more than 1 parallel request in this asynchronous queue? It's fine - just wondering!

packages/api-client/src/react-query/Event/useEvents.ts
```diff
@@ -89,6 +89,8 @@ export const useEvents = (
     },
     {
       staleTime: 5 * 60 * 1000,
+      refetchOnWindowFocus: false,
```
jimjeffers, line 92: Was this creating too many fetches?

jimjeffers (approved):
No issues here. I'm guessing this is now doubly fixed as it sounded like Carlos pushed a fix on the API to support blitzing the requests as we had been? Good to go!

packages/api-client/src/axios/Util/concurrencyLimiter.ts
ZackLyon: It was set to 3 concurrent requests, but it was still having the issue. That was before Carlos fixed it on his end, so it might be ok to do multiple now? It might be better to open a new issue to test it and get this one out the door

packages/api-client/src/react-query/Event/useEvents.ts
ZackLyon: In normal usage, it probably wouldn't, but during my testing, it was too easy to trigger refetches on these computationally heavy endpoints when the results don't actually change that often.

Issue #465, opened 2025-09-02
Critical - Cannot Override Mission Parameters.


ksalamy:
**Describe the bug**
Trying to override number of samples and cartridge type in sample_lab mission.
The Env observed was Maintenance/sample_lab.tl

**To Reproduce**
Cannot edit the number of samples and cartridge type when trying to override mission parameters

**Expected behavior**
Edits to the mission parameters should take effect.

ZackLyon:
Closed via #472

PR #472, opened 2025-09-04
Ensure parameter overrides work for all mission scheduling

ZackLyon:
- Clear overrides when switching mission selection categories (ie Recent Runs, Frequent Runs, Science), so that new overrides can take effect
- Updated default overrides to only be applied when selecting a previously run mission
- Update how plotted waypoints are defined so that waypoints will be reflected correctly in the final mission scheduling review step
- Map stations correctly so that latitude and longitude are being defined properly from the geojson data (the geojson array is longitude first, but that was previously being mapped to longitude). 

Screenshot 2025-09-04 at 1 12 48 PM


ZackLyon:
References #465

jimjeffers (approved):
This looks fine! Thanks for making this easy on the reviewer eyes.

Issue #466, opened 2025-09-02
The Vehicle Schedule does not reflect events outside of a deployment


ksalamy:
**Describe the bug**
Vehicle Schedule only shows current deployment information and not any other deployments.

**To Reproduce**
While deployed, open the schedule for the vehicle. The list only shows the current schedule for 
the vehicle.

**Expected behavior**
We expect to see the current schedule and and previous schedule and logs for the vehicle in its
schedule section. 

**Screenshots**
https://mbari.enterprise.slack.com/files/U04K1F8AP/F09CLEZ4XRR/img_6035.jpg

ZackLyon:
@ksalamy - I believe this is a repeat of the issue #429. If so, can we close this one since that one is still open? 
Friendly FYI: Unfortunately, I'm not able to see screenshots from the mbari slack channel, so you'll need to paste the actual image in for GH issues.

ZackLyon:
Closed via #479

Issue #429, opened 2025-05-09
Schedule history only goes back to the beginning of the last deployment


ZackLyon:
**Describe the bug**
Brian noted that the history in the scheduling section on the deployment page only goes back as far as the beginning of the last deployment, but it should allow you to go back further.

**To Reproduce**
1. Go to a vehicle deployment page
2. Click on the Schedule tab on the right hand side
3. Scroll through the history and see that it is truncated with no option to load more entries

**Expected behavior**
There should be an option to load more history entries beyond those from the last deployment

ZackLyon:
Update on this issue: This is currently using the `/deployments/commandStatus` endpoint, but the command statuses it sends back are all `TBD`. Contacted Carlos for more info since I couldn't find it in the Tethys api docs or changelog. Most likely this will need to switch over to using the events endpoint with the `run` and `command` event types and the status for each will need to be derived from comms/ACK, most likely.

ZackLyon:
Carlos has indicated that the `commandStatus` endpoint is not fully implemented. This issue is on hold until it is available.

heavyB:
Updating this issue to include schedule and comms queue as well. These need to be available outside of a deployment.

ZackLyon:
@heavyB - FYI: I've already updated comms queue to work outside of deployment in [GH#428](https://github.com/mbari-org/dash5/pull/428). It's been merged, but I don't know that it's been deployed to staging yet.

ZackLyon:
Closed via #479

PR #479, opened 2025-09-11
Add view historical logs option to schedule section

ZackLyon:
- Add toggle to switch between historical and deployment logs to schedule section
- Implement pagination features for historical logs including show more button
- Set up retrieval of event logs for historical logs section and format them to work with existing UI

References #429 and #466

### DIFF
--- a/apps/lrauv-dash2/components/ScheduleSection.tsx
+++ b/apps/lrauv-dash2/components/ScheduleSection.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef } from 'react'
+import React, { useState, useRef, useMemo } from 'react'
 import {
   AccessoryButton,
   AccordionCells,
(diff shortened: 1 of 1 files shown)


jimjeffers (approved):
I think this approach is sound though you could consider actually separating this into two different components that you swap out here. You're conditionally swapping two fairly different behaviors conditionally. So this page will load both the allLogsReponse and the deploymentResponse even though you're only viewing one or the other. It would be easier on the eye if you made two subcomponents that managed the networking logic for either use case and wrapped the existing chunk of UI they populate. This would conditionally call the appropriate network call for the exact situation and also make the component easier to read.

Not a deal breaker though. Up to you if you want to pursue this optimization or ship as is. The component is beefy but not totally out of hand yet.

ZackLyon:
That's a very good point, but since there are so many issues to tackle with this project, I think I'll leave it as is for now.

Issue #473, opened 2025-09-04
getUnitFromAbbreviation uses case-insensitive search, but some units are case-sensitive


sqshemet:
**Describe the bug**
Observed behavior:
1. Rerun a mission that overrides some timeout in hours (e.g. `set MissionTimeout 4 h`)
2. See that this populates the mission modal as "hour"
3. But on subsequent clickthrough becomes "henrys"

Pretty sure what's going on here is a bug, or at least an unintended side effect, in getUnitFromAbbreviation:

```export const getUnitFromAbbreviation = (
  abbreviation?: string,
  unitsData?: { abbreviation: string; name: string }[]
) => {
  if (!abbreviation || !unitsData) return undefined
  if (abbreviation.toLowerCase() === 'degree') return 'degree'
  return unitsData?.find(
    (u) => u.abbreviation?.toLowerCase() === (abbreviation ?? '').toLowerCase()
  )?.name
}
```

`(u) => u.abbreviation?.toLowerCase() === (abbreviation ?? '').toLowerCase()`
Is going to match the first case-insensitive unit that matches the abbreviation. But here's what the backend /get/units returns:
```    {
      "name": "henry",
      "abbreviation": "H"
    },
    {
      "name": "hertz",
      "abbreviation": "Hz"
    },
    {
      "name": "hour",
      "abbreviation": "h",
      "baseUnit": "second"
    },
```
So the lookup for case-insensitive 'h' gets matched to "henry" before "hour", and remains henry from that point forward.

Not sure how case-sensitivity is handled throughout the parsing process, so not submitting a PR with a suggested solution.

sqshemet:
Happy to write more detailed step-by-step instructions for an example that reproduces this if the problem isn't clear enough from the short description.

ZackLyon:
Closed via #480

PR #480, opened 2025-09-11
Make unit helper function case sensitive

ZackLyon:
- When looking up full unit names by abbreviation, use a case-sensitive search (ie: h = hour and H = Henry)

References #473


jimjeffers (approved):
Interesting issue. Goes against what I'm used to. Clean and easy though. Approved on my end Zack.

Issue #340, opened 2025-01-30
No track line shows for currently deployed vehicles when you click on the vehicles tab (not overview)


ksalamy:
When adding a currently deployed vehicle to Dash5, the vehicle and the track history show up in the OverView.

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/dd556817-3fc9-419b-8e93-4297d4a476f7" />
<br><hr><br>

However, neither the vehicle, its track lines, nor the map show up when the vehicle menu button is clicked.

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/929fb745-18ff-463a-a078-fd133036a5cb" />
<br><hr><br>

This is not the case for non-deployed vehicles (e.g., here Pontus). The last vehicle deployment shows up in the Vehicle OverView Page
similar to the currently deployed Vehicles:

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/68cc480f-0c15-4da8-aa79-b89e1626d493" />
<br><hr><br>

So does the Vehicle View Page when the vehicle menu button is clicked (!although displayed Off-Center!)

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/57fd1521-7a00-4e4b-a322-b959fae6488f" />

ZackLyon:
Closed via #481

PR #481, opened 2025-09-18
Show map and available track line for deployed vehicles

ZackLyon:
- Validate center coordinates and add default center so it is always available in Map to avoid cases where it causes map not to render
- Update VehiclePath to use **_either_** from/to props **_or_** lastDeployment start and end event times to use when querying the vehicle position endpoint to avoid random mismatches of start and end times
- Standardize the deployment start and end times on the deployment page to always use the start and end events from useCurrentDeployment instead of sometimes using useMissionStartedEvent to avoid random mismatches of start and end times
- Add bottom padding to the fit bounds for the deployment map in VehiclePath so that the vehicle path is always fully visible above the vehicle diagram on the deployment page
- Add optional chaining to many objects in the files listed above for objects that are potentially undefined to avoid property access errors on objects that don't exist
- Add comment to last deployment endpoint and hook that it sometimes returns future missions since that is not obvious from the name

### Currently deployed vehicle with available waypoints map
Screenshot 2025-09-18 at 3 24 14 PM

### Currently "deployed" vehicle with no waypoints
Screenshot 2025-09-18 at 3 07 59 PM

### Vehicle that is not deployed showing previous deployment on map
Screenshot 2025-09-18 at 3 07 16 PM

References #340

### DIFF
--- a/apps/lrauv-dash2/components/DeploymentMap.tsx
+++ b/apps/lrauv-dash2/components/DeploymentMap.tsx
@@ -141,26 +141,26 @@ const DeploymentMap: React.FC<DeploymentMapProps> = ({
   }, [vehicleName, setLatestGPS])
 
   useEffect(() => {
-    if (mapRef.current && !showVehicleColors) {
+    if (mapRef?.current && !showVehicleColors) {
       setTimeout(() => {
         try {
           // Try to invalidateSize method(s)
-          if (typeof mapRef.current.invalidateSize === 'function') {
-            mapRef.current.invalidateSize()
+          if (typeof mapRef?.current.invalidateSize === 'function') {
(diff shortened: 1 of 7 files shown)


jimjeffers (approved):
Sounds good. What about that comment? Is that a TODO? What should we do in the cases when that happens?

ZackLyon:
> Sounds good. What about that comment? Is that a TODO? What should we do in the cases when that happens?

It's just a friendly warning to those fools who might assume that the word last in last deployment indicates that it's in the past. One of the other endpoints was using the last deployment start time as its `from` and another value from somewhere else for its `to`, so occasionally it had back to the future time range.

PR #485, opened 2025-09-25
Fix partially loaded map

ZackLyon:
Weird issue where Leaflet will render some tiles as gray if it doesn't have a container with a non-zero height on load
- Add `min-h-0` to all flex parents of deployment map to resolve the issue
- Add styles that target the Allotment library panes to add minimum height zero to them

Note: it resolves the issue, so I'm not sure if it's worth looking for a more elegant solution since it's a tricky browser based issue

### Google's AI writeup on the issue
Screenshot 2025-09-25 at 1 29 21 PM


jimjeffers (approved):
Yeah that's fine. There are [some people who take issue ](https://tigerabrodi.blog/stop-using-min-heightwidth-for-flex-items-understand-flexbox) but most of us just use min-w or min-h hacks.

PR #486, opened 2025-09-30
Add text search for logs

### LINKED ISSUE #274: Request capability to type for a free text filter (similar as in dash4)
Deployment Daphne 03/15/2023:
Quinn -
I see how to filter by event type in the log, but not how to type for a free text filter as is possible in the old dash. Does that capability exist?

Also, Mission parameters should indicate or be grouped by their source behavior.

**Additional context**
These are feature requests not requirements.

ZackLyon:
- Create LogFiltersDropdown component with stories and tests that matches mockup design
- Implement text filtering on log results

Closes #274 

### Checkbox selection
Screenshot 2025-09-29 at 5 01 10 PM

### Text Filtering
Screenshot 2025-09-29 at 5 01 42 PM

### DIFF
--- a/apps/lrauv-dash2/components/LogsSection.tsx
+++ b/apps/lrauv-dash2/components/LogsSection.tsx
@@ -3,11 +3,12 @@ import { useInfiniteEvents, useTethysApiContext } from '@mbari/api-client'
 import {
   Virtualizer,
   LogCell,
-  MultiSelectField,
   AccordionCells,
   LogsToolbar,
   LoadMoreButton,
   SelectOption,
+  Button,
+  LogFiltersDropdown,
 } from '@mbari/react-ui'
 import toast from 'react-hot-toast'
 import { MultiValue } from 'react-select'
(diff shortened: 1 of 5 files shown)


apps/lrauv-dash2/components/LogsSection.tsx
```diff
@@ -99,15 +124,37 @@ const LogsSection: React.FC<LogsSectionProps> = ({
       const selectedFilterNames = filters.map(({ id }) => id)
       events = applyEventFilters(events, selectedFilterNames)
     }
+    if (debouncedSearchText.trim().length) {
+      const searchTerm = debouncedSearchText.toLowerCase()
```
jimjeffers, line 128: Wondering if debounce should be within the `LogFiltersDropdown` itself in case you need to reuse this in the future? Otherwise looks clean.

jimjeffers (approved):
Seems fine. Not might be worth packaging the debounce logic into the component actually managing search state for reusability but not a dealbreaker.

apps/lrauv-dash2/components/LogsSection.tsx
ZackLyon: This is a good idea. If I end up having to use it again, I'll do a quick refactor to add it.

PR #502, opened 2025-12-06
Rerun command from scheduler

ZackLyon:
- Verify whether a given schedule entry is actually a command or a mission (previously incorrect)
- Allow commands in scheduler to be rerun as freeform commands
- Next button is now active when rerunning from scheduler

Note: I tried for quite a while to allow commands that are rerun from the scheduler to use the actual command template builder with dropdown menus. However, due to the amount of variation in how commands are created and the inconsistent formatting, I was unable to get it to work across all commands. Hopefully using the freeform format will work for users.

## Selecting to rerun command
Screenshot 2025-12-05 at 5 02 45 PM

## Freeform command builder
Screenshot 2025-12-05 at 5 03 01 PM

## Another example of rerunning with freeform command builder
Screenshot 2025-12-05 at 5 03 42 PM

### DIFF
--- a/apps/lrauv-dash2/components/ScheduleSection.tsx
+++ b/apps/lrauv-dash2/components/ScheduleSection.tsx
@@ -58,6 +58,21 @@ export const parseMissionCommand = (name: string) => {
   }
 }
 
+/**
+ * Detects if a command is actually a mission command by checking if it contains
+ * "load" followed by a mission file path and "run"
+ */
+export const isMissionCommand = (
+  commandData?: string,
+  commandText?: string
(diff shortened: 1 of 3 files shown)


jimjeffers:
Ok I think freeform should be fine for the edge case. I remember the reason we avoid a freeform editor in this version is to prevent user errors in the command. But hopefully in this case this should be less likely as they are not typing the command from scratch.

jimjeffers (approved):
Looks good to me. Let's see if the team has any concerns with freeForm during testing.

PR #509, opened 2026-02-12
Display seconds in CommsSection timestamps (fixes #491)

### LINKED ISSUE #491: Display seconds in log
We'll often use vehicle log event timestamps at full resolution to correlate with other datastreams or sources of events. Minute resolution isn't always good enough to fully understand what's happening; seconds can be important.

I'd prefer to just always display seconds in the timestamps in the log, but if that's against prior consensus, maybe we could add the full-res timestamp as a mouseover tooltip or something?

ksalamy:
Extends the full-resolution timestamp display from LogsSection to CommsSection. Command and mission event timestamps now show H:mm:ss for correlation with other datastreams and sources.

Addresses #491.

### DIFF
--- a/apps/lrauv-dash2/components/CommsSection.test.tsx
+++ b/apps/lrauv-dash2/components/CommsSection.test.tsx
@@ -146,4 +146,21 @@ describe('CommsSection', () => {
       'text-indigo-600'
     )
   })
+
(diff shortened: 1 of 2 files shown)


ZackLyon (approved):
Looks good!

Issue #515, opened 2026-03-11
Refreshing the Browser causes the app to log the user out of their existing session as PIC or On-Call.


ksalamy:
**Describe the bug**
Refreshing the Browser causes the app to log the user out as PIC or On-Call. This always requires a log back in to the session.

**To Reproduce**
1. Go to 'dash5.mbari.org'
2. Click on the Vehicle name you will be watching for on top Nav. menu.
3. Click on the eye icon to the far right above the accordion menu for the vehicle.
4. Log in as PIC or On-call for the current vehicle you are on Watch for. Close with the Done button. 
5. Refresh the browser.
6. Notice that the eye icon now displays "unavailable" and requires the user to log back in before displaying they are on Watch again.**Screenshots**

**Desktop (please complete the following information):**
 - Browser [Tested on Chrome and Firefox]

ksalamy:
Fixed by PR #526 and PR #534, both merged into `develop`. The browser refresh logout was caused by two bugs: (1) `useRefreshSessionToken` was clearing the session cookie on successful token validation, and (2) `useSessionToken` had an SSR hydration issue in the static-export build. Both are resolved.

PR #526, opened 2026-03-12
fix(#515): fully resolve browser refresh logging user out of PIC/On-Call session

ksalamy:
## Root Causes & Fixes

Six bugs were found and fixed. The most critical was an SSR hydration cookie gap causing every browser refresh to appear as a full logout.

### 1. SSR hydration cookie gap (root cause)
`react-use-cookie` initialises via `useState()` on the server where `document.cookie` is unavailable, returning `''` every time. React reuses that server state during hydration — so the real cookie was never read on page load.

**Fix:** `useSessionToken` now initialises as `''` and syncs from `getCookie()` in a `useEffect` that only runs on the client, after hydration.

### 2. Cookie wiped on tokenless refresh response
The real okeanids `/user/token` endpoint returns 200 with profile data but **no `token` field**. The old `onSettled` handler resolved `data?.token ?? ''` to `''` and erased the session cookie on every refresh.

**Fix:** `useRefreshSessionToken` uses `onSuccess` with an explicit `if (data?.token)` guard — cookie is never cleared on a valid 200 response.

### 3. currentUser not restored after browser refresh
The `useEffect` hydrating `currentUser` required `refreshedSession.data.token` to be present. When the server omits the token, `currentUser` stayed `null` — leaving the user authenticated in the cookie but appearing logged out (avatar shows `uu`).

**Fix:** The effect falls back to the existing `sessionToken` from the cookie. The JWT payload is decoded to recover `firstName`/`lastName`/`email`/`roles` so the avatar shows real initials instead of `uu`.

### 4. PIC/On-Call `Unavailable` flash during refresh
A one-render gap between `refreshedSession.isLoading` going false and `setCurrentUser` being called caused `authenticated` to briefly be false, making `createRoleLabel` return `Unavailable`.

**Fix:** A `pendingAuthValidation` flag keeps `loading: true` until `currentUser` is hydrated.

### 5. Expired token not surfaced
When the server rejected an expired token (401), the user was left in a broken authenticated state.

**Fix:** A `useEffect` calls `logout()` when `refreshedSession.isError` is true and the user had an active session.

### 6. PIC/On-Call query gated on auth unnecessarily
An earlier incomplete fix added `enabled: hasToken` to `useVehiclePicAndOnCall`. The events endpoint is public — this gate blocked PIC display during the refresh window.

**Fix:** Removed the gate; auth headers still sent when available; query key still differentiates authed vs unauthed.

## Test Plan
- [ ] Log in, refresh browser — stay logged in with correct initials (not `uu`)
- [ ] Log in as PIC for a vehicle, refresh — PIC label preserved, no Unavailable flash
- [ ] Simulate network timeout during token refresh (DevTools throttle) — not logged out
- [ ] Replace cookie with an expired JWT — login prompt shown cleanly


ksalamy:
CI is green, Copilot feedback has been fully addressed, and the branch now correctly clears in-memory auth state when refresh tokens are invalidated while preserving session behavior for transient refresh failures.

ksalamy:
CI is passing, Copilot follow-up comments have been fully resolved, and the PR now preserves session continuity on transient refresh failures while correctly clearing stale auth state when refresh returns no valid token.

ksalamy:
Hey Zach,

PR #527 has been closed. The complete fix is now in #526 only — that's the only one you need to review.

The unrelated tiptap/Doc section code has been removed and will go through its own PR separately.

The fix has been tested locally — session persistence, user initials, and PIC/On-Call display are all working correctly on refresh.

Thanks

packages/api-client/src/react-query/TethysApiProvider/TethysApiProvider.tsx
```diff
@@ -107,6 +140,28 @@ export const TethysApiProvider: React.FC<TethysApiProviderProps> = ({
     onSessionEnd?.()
   }, [setCurrentUser, setError, setLoggedOut, onSessionEnd])
 
+  // When the server rejects the session token (e.g. expired JWT → 401), clear
+  // auth state so the login prompt is shown rather than a broken "authenticated"
+  // UI with an expired token. Only fires when the user was actively logged in
+  // (existingToken is truthy) to avoid interfering with the initial page load
+  // path where the user was never authenticated.
+  useEffect(() => {
+    if (refreshedSession.isError && existingToken && !loggedOut) {
```
ZackLyon: The AI review I did pointed out that refreshedSession.isError is true for any error code, not just a bad or expired token. You could consider changing this to explicitly catch the invalid auth code. Leaving it the way it is will be fine, but it will log the user out if there's any kind of server error.

apps/lrauv-dash2/package.json
```diff
@@ -20,23 +20,25 @@
...
-    "@tiptap/extension-table-cell": "^3.13.0",
-    "@tiptap/extension-table-header": "^3.13.0",
-    "@tiptap/extension-table-row": "^3.13.0",
-    "@tiptap/extension-text-align": "^3.13.0",
-    "@tiptap/extension-text-style": "^3.13.0",
-    "@tiptap/extension-underline": "^3.13.0",
-    "@tiptap/pm": "^3.13.0",
-    "@tiptap/react": "^3.13.0",
-    "@tiptap/starter-kit": "^3.13.0",
+    "@tiptap/core": "3.13.0",
```
ZackLyon, line 23: All the tiptap code is still here. It's fine if you don't want to bother removing it.

apps/lrauv-dash2/components/docs/DocFormEditor.tsx
```diff
@@ -315,13 +318,18 @@ export default function DocFormEditor(props: DocFormEditorProps) {
       TextAreaNode,
     ],
     content: contentWithIds || '',
-    onUpdate: ({ editor }) => {
+    onUpdate: ({ editor }: { editor: Editor }) => {
```
ZackLyon, line 321: Let me know what I should be looking for or testing if you want to leave the tiptap code in this PR.

packages/api-client/src/react-query/TethysApiProvider/TethysApiProvider.tsx
```diff
@@ -65,20 +88,30 @@ export const TethysApiProvider: React.FC<TethysApiProviderProps> = ({
...
+        const apiProfile = refreshedSession.data
+        const profile = apiProfile.firstName
+          ? apiProfile
+          : decodeJWTProfile(token)
+        setCurrentUser({ ...profile, ...apiProfile, token })
+      }
     }
   }, [
     refreshedSession.data,
     lastInitialToken,
```
ZackLyon: `lastInitialToken` is a ref that is never set to a value, but is then used as a dependency for this useEffect. We should remove it since it seems like it's leftover dead code.

ZackLyon (changes requested):
We should definitely remove lastInitialToken in TethysApiProvider and look over the other things I noted and decide if you want to make any of those changes. I checked the docs section really quick and it _seems_ fine. Not sure if you want to leave those changes in here?

ksalamy:
The tiptap change in `DocFormEditor.tsx` is purely a TypeScript type annotation — adding `{ editor: Editor }` to the `onUpdate` callback so it compiles cleanly against the updated tiptap version. There's no functional change at runtime. To spot-check: open any vehicle's Docs section, open a note in the editor, and confirm you can type and format normally. That's all there is to it.

ksalamy:
Hi @ZackLyon, addressed your feedback — here's what was updated:

- **Removed `lastInitialToken`** — the ref was never written to and was dead code sitting in the `useEffect` dependency array. Cleaned up.
- **Tightened the `isError` logout** — the `useEffect` that clears auth state now only fires on explicit **401/403** responses, not transient server errors (500, network timeouts, etc.). That way a brief server hiccup won't log a user out of an otherwise valid session.

Leaving the tiptap code as-is since it's a TypeScript-only annotation fix with no runtime impact (as noted in the earlier comment).

Ready for another look when you get a chance — thanks!

ZackLyon (approved):
Thanks for making those changes! Approved.

ksalamy:
fix(#515): fully resolve browser refresh logging user out of PIC/On-Call session
- Preserve session token on transient refresh errors; only clear on
  explicit 401/403 auth rejections to avoid logging out on server hiccups
- Decode JWT payload as fallback when API response omits profile fields,
  fixing "uu" initials after refresh
- Restore PIC/On-Call status after browser refresh by passing auth headers
  in useVehiclePicAndOnCall requests
- Remove dead lastInitialToken ref from TethysApiProvider
- Update useRefreshSessionToken tests to match corrected behavior

PR #530, opened 2026-03-16
feat: add schedule history details modal quick view

### LINKED ISSUE #517: Would like a popup of full Mission parameters when clicking on a Scheduled Mission (current of historical)
There is currently no way to see what the current or past vehicle deployment parameters are unless you click on an old one an temporarily "use it for new Mission". At that point one can see the settings but it quite hard to just see what the mission parameters are for the user. Often you have to scroll back through the logs history just to find the mission calls. 

It would be much more efficient and helpful to the users to simply click on the mission and have a modal display all relevant information quickly to the user.

ksalamy:
### Core feature: Schedule Event Details modal
- Adds a draggable, non-blocking Schedule Event Details modal that opens when clicking any schedule history mission or command row
- Displays key read-only metadata: mission/command label, status, scheduled time, started time, ended time, operator, note, transmission mode, and parsed command segments
- Status and Ended help tooltips (keyboard/mouse/focus accessible via `aria-describedby` + `role=tooltip`) explain the approximate nature of lifecycle data to operators
- Modal header title derives from event type: "Mission Details" vs "Command Details"
- Extends shared Modal header styling hooks for custom header/title/close-button styling
- CopyButton clears its timeout ref on unmount to prevent state updates on unmounted components

### Schedule row status improvements
The API's `/deployments/commandStatus` endpoint returns `TBD` for all row statuses and empty `relatedEvents` arrays — there is no server-side lifecycle signal beyond what the vehicle logs. Status accuracy is therefore derived from the `/events/mission-started` timeline:

- **Currently running mission**: detected from `mission-started[0]` (newest-first); polling tightened from 60s to 30s interval, stale threshold 30s to 15s for faster running-to-ended transitions
- **Mission matching**: reducer skips already-completed rows and selects the most recently sent command; a 30-minute satellite delay window handles `sched asap` transitions that arrive after the vehicle has already started the mission
- **Stale pending promotion**: any row whose reference time predates the currently running mission's start is promoted to `completed`; `isRecovered` cleanup now handles both `pending` and `TBD` raw API statuses
- **Comms-derived status**: `useCommsEvents` SBD events promote eligible rows to `sent`, `ack` (Received), or `timeout`; polling every 30s for live updates
- **Status normalization**: `toScheduleCellStatus` maps `TBD` and unknown values to `pending`; `scheduleStatus` gate and modal meta use the normalized `cellStatus` consistently

### Default Mission support
- Autonomous Default missions (vehicle fallback when no operator mission is running) are now shown as first-class rows in Schedule History
- Synthetic rows derived from gaps in the `mission-started` timeline; filtered to >60s duration to exclude noise; carry GPS `fix` data for context
- Dedicated Default Mission modal: light-green header, deployment-prefixed "Mission Details" title, conditional duration text ("Default mission has been running for X" / "Default mission ran for X"), italic descriptive text, and a Status tooltip explaining Running vs Completed semantics
- History is sorted chronologically by `unixTime` descending so Default rows appear in the correct position

### Icons and visual polish
- Custom SVG `ConnectedIcon` (sent to vehicle) and `AcknowledgeIcon` (received by vehicle) replace FontAwesome paper airplane icons for `sent` and `ack` statuses — consistent with the Comms queue icons
- Config update tooltip changed from "added to running mission" to "sent to vehicle" for accuracy
- Separator between active and historic rows changed from "Completed & Earlier" to "Previous Vehicle Directives"
- Schedule History header note: "End times are approximate / See Logs for exact times" (two lines)
- Precise relative timestamps: custom `timeAgo()` replaces Luxon's `toRelative()` for "Xh Ym ago" granularity
- Schedule row labels: teal-green (`text-teal-600`) for all rows, blue reserved for the running mission; pending clock uses regular (not solid) icon; flat white backgrounds on all rows

### OverviewToolbar: Recovered pill
- Added `recovered` and `recoveredAt` props to `OverviewToolbarProps`
- Renders a green "Recovered" pill inline next to the deployment name with a tooltip showing the recovery timestamp

### Leaflet / GoogleMutant crash fix
- Patches `L.GridLayer.GoogleMutant.prototype._setupAttribution` with a null guard on `this._map?._controlCorners` to prevent a `TypeError` when the map unmounts during an asynchronous Google Maps API callback

Closes #517

## Test plan
- [x] Click schedule history mission/command rows — verify modal opens with correct details
- [x] Click another row while modal is open — verify contents update
- [x] Verify Status and Ended tooltips appear on hover/focus/click
- [x] Verify modal header shows "Mission Details" for missions and "Command Details" for commands
- [x] Verify rows older than the current running mission no longer show as pending after ~30s
- [x] Verify future `sched YYYYMMDDTHHMM` rows remain pending if their scheduled time is after the current mission started
- [x] Verify Default mission rows appear in chronological order in Schedule History
- [x] Verify Default Mission modal shows green header, correct duration text, and status tooltip
- [x] Verify `ack` rows show ConnectedIcon / AcknowledgeIcon and "Received by <vehicle>" tooltip
- [x] Verify `timeout` rows display correctly
- [x] Verify modal is draggable and close button works
- [x] Verify lint and all unit tests pass (lrauv-dash2: 52/52, react-ui: 479/479, api-client: 311/311)


ksalamy:
Hi @ZackLyon,

PR#530 is a big Schedule Section update. As a result this comment will be a bit long to describe the logic behind the requested changes. I had a lot of help from Claude with this PR as far as logic and function of that logic throughout these updates. Our logic was solidly based in the logs from Dash4.  - Karen and Claude

--

The Schedule History section existed before this PR, but every row showed the same placeholder status from the API (TBD) with no way to tell operators what had actually happened. Over the past month — drawing on direct conversations with team members, real Dash4 production logs, and PIC experience in the field — we worked through every row type to understand what we can accurately determine and what we cannot. This PR reflects that audit. Where we are confident, we advance the status. Where we are not, we say so.

**The data problem**
The /deployments/commandStatus API endpoint returns TBD for every row. There is no server-side lifecycle signal — the vehicle does not report back a neat "mission complete" flag. Everything shown in the Schedule History is therefore assembled from three independent sources:

- mission-started telemetry — emitted by the vehicle itself when it begins executing a mission. This is the highest-confidence source we have.
- SBD comms events — the satellite layer's record of whether a transmission was sent and whether the vehicle's modem acknowledged receipt.
- The operator command log — what was commanded, by whom, when, and via which channel.
- None of these alone is sufficient. The status logic combines all three.

**What each status means and how it is derived**
Running Sourced from mission-started[0] — the newest entry in the vehicle's own telemetry stream. The vehicle emits this event (with GPS fix and mission name) the moment it begins executing. Index 0 is always the currently active mission. **This is the closest thing we have to real-time ground truth.** The Schedule History polls this endpoint every 30 seconds with a 15-second stale threshold so the transition from Running → Ended surfaces promptly.

- Completed - A mission is considered completed when a newer mission-started event has replaced it. The ended time is set to the moment the next mission began — inferred, not directly reported. We do not call a mission Completed until we have positive evidence of a successor. Approximation markers (~) appear throughout the modal UI to signal this.
- Sent - The default status for any non-mission command once it leaves the operator's queue. It means: the command was transmitted; no vehicle acknowledgment has been received yet.
- Received - Upgraded from Sent only when the SBD comms layer returns an ack for that specific event ID. This means the vehicle's modem confirmed receipt of the transmission. It does not guarantee the vehicle executed the command — that distinction is made explicit in the row tooltip ("Received by <vehicle>").
- Timeout - Upgraded from Sent when the comms layer returns a timeout for that event ID — the satellite transmission was never acknowledged. Shown honestly. The vehicle may or may not have received the command.
- Pending - Used only for future-scheduled missions (sched YYYYMMDDTHHMM). We use the scheduled run time from the command string as the matching reference point — not the send time — so a mission queued today for tomorrow is never incorrectly promoted.

**How mission matching works**
Matching a command row to a mission-started telemetry entry is complicated by satellite latency. A sched asap command can be sent while the old mission is still running, but the vehicle won't receive and execute it until 10–30 minutes later. The algorithm handles this with three layers:

- Interval containment — if the command's reference time falls within [startedAt, endedAt) for a timeline entry, that entry is the direct match.
- Satellite delay window — if interval containment lands on a completed entry but a newer run of the same mission begins within 30 minutes of the command send time, the match is redirected to that newer run. This correctly handles sched asap transition commands.
- Nearest-by-time fallback — if neither of the above applies, the closest timeline entry within a 10-minute window is used. Anything beyond 10 minutes is left unmatched to avoid false positives.
- When the same mission is commanded more than once, the most recently sent command is treated as the active one — not the one whose send time happens to be numerically closest to the vehicle's start time.

**Default Missions**
Default missions are the vehicle's autonomous fallback — not operator-commanded. They run when no mission is active (after a timeout, a stop command, or a cold boot). Per the project team's own definition: "it's basically just what to run if nothing else is commanded."

Default mission rows are derived entirely from the mission-started timeline: any entry where the mission name normalizes to "default" is treated as a Default mission. Start time comes from the vehicle's own telemetry (high confidence). End time is inferred as the moment the next operator mission began, marked with ~. Runs shorter than 60 seconds are suppressed as inter-mission noise.

The Default Mission modal makes clear in its description that this is an automatic fallback, not an operator-directed action.

**Schedule Event Details Modals**
Each row in the Schedule History opens a details modal on click. The modal displays the same derived status as the row, along with:

- Started — the vehicle's actual execution start time from mission-started telemetry (for missions), or the command send time (for commands). Shown as both an absolute timestamp and a relative "Xh Ym ago" for quick context.
- Ended — inferred end time (marked ~) for missions; N/A for instantaneous commands (parameter updates, one-off commands) since they have no meaningful duration.
- Status tooltip — a ? button on the Status field explains what each status means in plain language, including its accuracy limitations.
- Default Mission modal — uses a distinct green header and deployment-prefixed title to make clear this was an automatic fallback, not an operator dispatch. Duration text is conditional: "Default mission has been running for X" while active, "Default mission ran for X" once ended.
- Relative timestamps — replaced Luxon's broad toRelative() ("3 hours ago") with a custom timeAgo() that shows "3h 22m ago" for better precision given that approximated end times already carry some uncertainty.

**What we do not claim**
To be precise about the boundaries of this work:

- End times for all missions are inferred — they represent when the next mission began, not when the vehicle reported completion. All are marked ~.
- Completed - means the vehicle stopped running that mission and started something else. It says nothing about whether the mission achieved its scientific objectives.
- Received - means the vehicle's modem confirmed the SBD transmission. It does not confirm the command was executed.
- We do not distinguish between a mission that timed out and one that was manually stopped. Both appear as Completed (ended) because no log signal separates them.

This is consistent with operator expectations and was cross-validated against Dash4 production logs during development.

ksalamy:
Hi @ZackLyon,

Heads up: PR #531 was merged to develop ahead of this one. There are conflicts in ScheduleSection.tsx, ScheduleCell.tsx, missionUtils.ts, and their tests — both PRs made additive changes to the same files. Resolution should preserve both sets of changes. 

If you want, I could also proactively resolve the conflicts now by rebasing the #530 branch onto current develop. That would make it a clean merge for you with no conflict markers to deal with. Let me know if you would like me to do that. 

Sorry for the confusion. I should have caught that.

Karen

ZackLyon (approved):
@ksalamy - This is looking good overall. The one refactoring issue you might want to address is the length of the `ScheduleSection` (1159 lines) and the `ScheduleEventDetailsModal` (892 lines).
As an additional PR note: it appears that a bunch of unrelated tests got modified or added, as well as unrelated DocsSection changes. I assume the tests are meant to improve these files, but it's unnecessary extra noise when reviewing a large PR like this.

ksalamy:
Hi @ZackLyon:

**Decision:**  Merging now — follow-up refactoring tracked in two new issues.

The feature is working and approved. Per Zack's review, ScheduleSection.tsx (~1,159 lines) and ScheduleEventDetailsModal.tsx (~891 lines) have grown large enough to warrant splitting into focused sub-components. Rather than delay this PR further, I am merging as-is and tracking the refactors separately:

**Issue to open:** Refactor ScheduleSection.tsx — extract sub-components to bring the file under ~400 lines
**Issue to open:** Refactor ScheduleEventDetailsModal.tsx — extract modal steps/sections into focused child components

The unrelated test improvements and DocsSection touch noted in the review are intentional small improvements and not being reverted. Sorry for any noise but I still feel they are warranted. Thanks.

Issue #488, opened 2025-10-01
Cancelled requests shown as pending


sqshemet:
**Describe the bug**
Cell requests can be cancelled if they haven't been acked by the vehicle or timed out. Cancelled requests are shown as "waiting to transmit" in the comms queue.

**To Reproduce**
1. Send a cell-only command
2. Immediately cancel it

**Expected behavior**
Shown in comms queue as cancelled

**Screenshots**

<img width="543" height="110" alt="Image" src="https://github.com/user-attachments/assets/c4757047-1376-45d4-806c-4348e54b2e3f" />

<img width="1152" height="203" alt="Image" src="https://github.com/user-attachments/assets/39286f0c-a093-4e3b-aa79-617406098ab0" />

ksalamy:
Resolved in develop — Cancelled mission/command status is now correctly shown as 'Cancelled' in the Comms Queue (PR #567: fix/cancelled-mission-status-incorrect). Further improvements to status accuracy for mission commands also landed in PR #590 (fix/#584).

PR #528, opened 2026-03-12
fix(#514): show full document name and docType in Docs section

ksalamy:
## Summary

Documents with similar or identical names (e.g. three separate docs all labeled "Ahi-42 - MBTS") were indistinguishable in the Docs section because:

1. `DocCell` truncated document names to 3 words, hiding any differentiating suffix
2. The `docType` field (`NORMAL`, `FORM`, `TEMPLATE`, `FILLED`) was fetched but never shown

### `packages/react-ui/src/Cells/DocCell.tsx`
- Removed the 3-word truncation logic — the full `label` is now shown
- CSS `truncate` class already on the button handles overflow safely
- Added `title={label}` tooltip so the full name is always accessible on hover
- Added optional `secondary?: string` prop rendered as a small muted label below the document name

### `apps/lrauv-dash2/components/DocsSection.tsx`
- Passes `secondary={item?.docType}` to `DocCell` so the document type appears beneath the title (displayed lowercase: `form`, `normal`, `template`, `filled`)

## Before / After

**Before:** Three docs all showing `Ahi-42 - MBTS` with no way to tell them apart

**After:** Each doc shows its full name + doc type label below (e.g. `form`, `normal`). Deployment/vehicle attachment tags below still show which deployment each doc belongs to.

## Note on retroactive data changes

Document names in the database are set by users at creation time — this fix does not change any stored data. Users who want more descriptive names can rename individual docs via the Edit option. The `docType` field is already stored server-side and requires no data migration.

## Test plan

- [ ] Open Docs section for a vehicle with multiple similarly-named documents — names should now be fully visible
- [ ] Verify `docType` label (e.g. `form`) appears beneath each document title
- [ ] Verify long document names truncate gracefully and show full name on hover (title tooltip)
- [ ] Verify existing attachment chips (deployment/vehicle tags) still appear below

PR #531, opened 2026-03-16
fix: improve schedule running-status accuracy and matching safety

### LINKED ISSUE #516: Schedule History displays a current deployment as Ended when it is in reality still running.
**Describe the bug**
The Schedule History erroneously displays that a new mission on a current deployment (known to be currently running) has ended. This is a bug and the code logic needs to be investigated.

**To Reproduce**
1. Go to 'dash5.mbari.org'
2. Click on the Vehicle name in top navigation menu once it is added.
3. Click on the Schedule accordion that appears to the right. 
4. Note that the Schedule Header (blue) is different in the timing and status of the mission than the Schedule history statements.

**Expected behavior**
The Schedule History for a mission should not display ended until the mission has in fact ended. It should also sync up with the Schedule Header statement that is also listed. 

**Screenshots**
Image

ksalamy:
## Summary
- improve Schedule History running/completed resolution by using mission-started interval containment and safer fallback matching
- harden matching logic with path-aware comparisons (when available) and a guard against empty mission keys to reduce false running indicators
- expand mission utility tests for extensionless load commands, path normalization, and mission path extraction

Closes #516
Related to #488

## Test plan
- [x] Run `yarn workspace @mbari/lrauv-dash2 test:ci apps/lrauv-dash2/jest/missionUtils.test.ts`
- [x] Verify newest active mission appears in running section under Schedule History filter row
- [x] Verify completed missions remain in historical section with approximate end times
- [x] Verify mission matching works for load commands with and without file extensions
- [x] Verify lint passes for changed files


ksalamy:
Hi @ZackLyon 

This PR improves schedule running/completed status accuracy in the Schedule section. 

- The core change derives "running vs. completed" from the /events/mission-started timeline rather than relying solely on the raw command-status field, which was frequently stale or incorrect. 

- Mission matching was also hardened to handle extensionless load commands (e.g. load Default;run) and paths with dashes/dots. 

- A few supporting improvements came along for the ride: fitBounds is now wrapped in try/catch to prevent Leaflet crashes, DocCell gained a secondary label for doc type, and the elapsed time in the Schedule section now shows precise minutes (e.g. 1h 25m ago) instead of rounding to whole hours. 

All Copilot review rounds have been addressed. 

Thanks for taking a look!
Karen

ZackLyon (changes requested):
This is looking great overall. It certainly looks like it will make the Schedule section more helpful.

One thing to change:
The elapsed time since the mission started is only shown in hours and minutes, but it would much better to show it including days, months, etc. Some of them show thousands of hours. There's already some util functions to take care of this, so probably choose one of those and extend it if it doesn't quite fit: formatCompactDuration or formatElapsedTime would probably work. I think currently formatCompactDuration caps itself at 6 days (it shows ">6 days" for anything over that duration), but you could add an additional maxDuration prop so when that is undefined, it does not get capped (just make sure to backfill the places where we want the 6 day cap).

Image

A couple notes:
There is a bunch of unrelated code in this one also, so consider removing it. It's important to be working on different branches so your work doesn't end up overlapping. I usually create multiple instances of the repo on my computer and work on one branch in each of them if that's helpful.

Some of the missionUtils functions are similar to other util functions already in use. They are slightly different, so I think it's ok to keep them, but consider checking for existing utils in the future that you can modify or extend to keep things DRY.

ksalamy:
Hi @ZackLyon — all of your review comments have been addressed. Here's a summary of everything done since your last pass:

Your requested changes:

✅ Elapsed time now shows full years/months/days — extended formatCompactDuration with an optional maxDays param; when undefined, durations show as 4mo 2d, 1y 3mo, etc. instead of truncating at 6 days. All existing call sites (comms age displays) explicitly opt in to the 6-day cap.
✅ Unrelated code removed — stripped changes belonging to other PRs (#528 docs, #543 map fixes, unrelated tests) to keep this branch focused

**Additional improvements made during Copilot review (10 rounds):**

- formatCompactDuration: days round up when remainder hours ≥ 12 (e.g. 6d 23h → 7d); >1 day vs >N days pluralization; long-duration branch only activates when maxDays is undefined
- ScheduleSection: schedule cell description now uses status-aware labels (Queued/Scheduled for pending, Started for running, Ran for completed/cancelled); (ago) suffix suppressed for queued/pending items
- ScheduleSection: endedAt now derived from mission timeline for all rows regardless of existing API status (was previously skipped for non-TBD items, leaving description2 empty for historical items)
- ScheduleSection: mission-started events sorted by unixTime descending before building the timeline (defensive against non-guaranteed API ordering)
- ScheduleSection: missionStartedResponse.data removed from missions useMemo deps — missionTimeline entries now carry eventId and text, making it the single dependency
- ScheduleSection: DateTime.now() captured once before JSX return for consistent per-render timestamps
- Synthetic running-row injection: no longer invents an extension — uses bare mission name as-is to avoid breaking MissionModal auto-selection for .xml missions
- VehicleInfoCell / VehicleDiagram: DateTime.now() captured once per render; day comparison uses hasSame(now, 'day') instead of day-of-month comparison (was broken across month boundaries)
- useVehicleInfo: refetchInterval now wired to the primary query, not just the fallback
- UseVehicleInfoOptions: redundant refetchInterval redeclaration removed (inherited from SupportedQueryOptions)
- SupportedQueryOptions: refetchInterval?: number | false added
- Tests: 5 focused component tests added covering running status enrichment, synthetic row injection, completed classification, and null text safety; MSW mocks corrected to use /events (the actual data source when deploymentLogsOnly is false) and proper /deployments/commandStatus response shape; jest config updated with transformIgnorePatterns for react-dnd ESM and ResizeObserver mock

Thank you for the thorough review, Zach — the Copilot rounds were extensive but the codebase is meaningfully better for it. Appreciate your patience!

Karen

ksalamy:
Hi @ZackLyon,

Thanks for the feedback. 

**The following were addressed since your last review:**

- Elapsed time now uses formatElapsedTime from @mbari/utils, so long-running missions correctly display days, months, and years instead of just hours and minutes (e.g., 98d:12h ago instead of 2340h ago).
- Schedule header text is now dynamic — it reads "Schedule is paused" or "Schedule is running" to stay consistent with the background color and button label.
- useVehicleInfo spread order fixed so the enabled guard on !!params.name can't be silently overwritten by supportedOptions.

Copilot gave it a clean pass (no new comments) after these changes. Ready for your review!

-Karen

ZackLyon (approved):
@ksalamy  - Looks great! Ship it!

PR #532, opened 2026-03-20
GH459 Update Frequent Runs in Mission Scheduling to Correctly Apply Parameters

### LINKED ISSUE #459: Mission Scheduling with Frequent Runs Does Not Apply Parameters
**Describe the bug**
When scheduling a mission and selecting from the frequent runs options, the selected run will be the original run without any parameter overrides (ie if it's a run that loads sci2/flat_and_level then has a bunch of parameter overrides applied, it will just look like the original sci2/flat_and_level).

**To Reproduce**
1. Go to mission selection
2. Select an option under frequent runs
3. See that the parameter/waypoint overrides that are part of that mission are not applied

**Expected behavior**
Selecting a frequent run should apply all parameter and waypoint overrides associated with it

**Additional context**
Recent runs has already been updated to address this same issue

#### Frequent Runs for Sim on dash5
Image

#### Frequent Runs for Sim on dash4
Image

ZackLyon:
Closes #459 

- **Frequent runs** now **parse the full written command** (same style as recent runs): parameter and waypoint overrides are read from the API string, including **optional script metadata** when lat/lon-like names need to be classified as waypoints vs parameters.
- **Mission list entries** for frequent runs carry **override counts**, **override payloads**, and when there are overrides a **stable composite id** so two runs of the same template with different overrides stay distinct; **`missionPath`** always points at the real template path for loading the script.
- **Script loading** uses **`missionPath`** when the selected row is a composite id, so the correct mission definition (and insert/arg layout) is used in the modal.
- **Mission modal** applies **parameter, comms, safety, and waypoint** defaults for **Frequent Runs** the same way it does for **Recent Runs**, and **scheduling** / command creation uses the **template path**, not the composite id.
- **Schedule-history “rerun”** temp missions now set **`missionPath`** so path resolution stays consistent.
- **`useParameterOverrides`** takes a combined **recent + frequent** list and applies overrides when the category is either Recent or Frequent.
- **Shared helper** `extractOverridesWithScriptMetadata` centralizes “extract overrides, then refine with script lat/lon pairs when needed”; **recent runs** use it instead of duplicating that logic.
- **`generateMissionKey`** is **exported** from `@mbari/api-client` so frequent runs use the **same key algorithm** as recent runs.
- **Mission modal parameters**: **`useManagedParameters`** can **re-sync** `defaultOverrides` when they arrive **asynchronously** (e.g. after frequent-run processing), without overwriting the map once the pilot has started editing.
- **`Mission` type** gains optional **`missionPath`** for selections where **`id`** is not the template path.


jimjeffers (dismissed):
The big issue is useFrequentRuns is being fixed with a band-aid here. The api-client has abstractions for all of our API requests but you are consuming the hooks in the client and then attempting to merge state there. It'd be better just to resolve this directly in the API client via a dedicated hook:

```
export interface NormalizedMissionRun {
  selectionId: string
  missionPath: string
  writtenCommand: string
  waypointOverrides: Override[]
  parameterOverrides: Override[]
  waypointCount: number
  parameterCount: number
}

export const useFrequentRuns = (
  params: GetFrequentRunsParams,
  options?: SupportedQueryOptions
) => {
  const { axiosInstance, token } = useTethysApiContext()

  return useQuery(
    ['command', 'commands', 'frequent', 'runs', params.vehicle, params.limit],
    async () => {
      const commands = await getFrequentRuns(params, { instance: axiosInstance })
      const scriptRequestByMission = new Map<string, Promise<GetScriptResponse>>()

      const getScriptMetadata = (missionPath: string) => {
        let request = scriptRequestByMission.get(missionPath)
        if (!request) {
          request = getScript(
            { path: missionPath, gitRef: 'master' },
            {
              instance: axiosInstance,
              headers: { Authorization: `Bearer ${token}` },
            }
          )
          scriptRequestByMission.set(missionPath, request)
        }
        return request
      }

      const runs = await Promise.all(
        commands.map(async (writtenCommand) => {
          const missionPath = parseMissionPathFromCommand(writtenCommand)
          if (!missionPath) return undefined

          const { waypointOverrides, parameterOverrides } =
            await extractOverridesWithScriptMetadata(
              writtenCommand,
              missionPath,
              () => getScriptMetadata(missionPath),
              { logContext: missionPath }
            )

          return {
            selectionId: generateMissionKey({
              missionName: missionPath,
              waypointOverrides,
              parameterOverrides,
            }),
            missionPath,
            writtenCommand,
            waypointOverrides,
            parameterOverrides,
            waypointCount: waypointOverrides.length,
            parameterCount: parameterOverrides.length,
          }
        })
      )

      return runs.filter(Boolean)
    },
    {
      staleTime: 60 * 1000,
      ...options,
    }
  )
}
```

I guess the key questoin is whether or not the requirements for frequent runs are supposed to be scoped to the vehicle or across all vehicles. IF it's meant to be all vehicles then you should ignore the advice on ensuring the cacheKey is normalized to the current vehicle name / id. We're not currently doing that in the existing implementation so if you switch vehicles you will hit the cache when you come back to the modal using this hook unless it's stale.

Here are all 3 findings that codex found for me. The first is top priority but you should consider the other two as well:

>useFrequentRuns is still the wrong abstraction boundary, so the PR fixes the symptom in app code instead of fixing the source model. Today the hook only returns a parsed { writtenCommand, mission } wrapper and uses a static React Query key, which means vehicle-specific frequent runs are not part of the cache key at all: useFrequentRuns.ts (line 6). The PR then adds a second useQuery in app code to fetch script metadata, classify overrides, and generate composite ids. That is workable, but it spreads “what is a frequent run” across two layers instead of making the API hook return a normalized run object.

> The hasUserModifiedRef change in PATCH 3 looks prone to stale override state across mission switches. In the PR, async defaultOverrides are only re-synced when that ref is false, and resetOverrides is added but I did not see it wired into mission selection flow. The current modal consumer just calls useManagedParameters(...) and updates selectedMissionId on selection: MissionModalView.tsx (line 153), MissionModalView.tsx (line 203). That makes this a likely regression if a user edits one mission, then selects another frequent run before closing the modal.

> The PR’s missionPath addition is a pragmatic escape hatch, but it confirms the model is overloaded. The current flow assumes the selected row id is also the script path for both loading and command creation: MissionModal.tsx (line 264), MissionModal.tsx (line 355). The PR patches each call site by saying “sometimes id is a composite run key, sometimes it is a real mission path.” That works, but it is not the cleanest contract.

ZackLyon:
Refactored to address the concerns mentioned above. Also included functionality to allow the user to optionally use frequent runs for all vehicles using the checkbox as was already present for recent runs.

ZackLyon:
@ksalamy - Can you please review and verify that I have addressed all of Jim's review requests?

ksalamy (approved):
Hi @ZackLyon  — great work on this refactor. I reviewed the code and tested the Frequent Runs flow up through the scheduling step on a live vehicle deployment. Everything looks solid.

Going through Jim's three review requests:

1. useFrequentRuns abstraction boundary — Fully addressed. All the override extraction, script metadata fetching, and key generation logic now lives inside the hook itself. The NormalizedMissionRun type matches Jim's suggested interface (plus the added vehicle field for multi-vehicle support). The cache key is properly normalized to vehicleKey + gitRef + limit so cross-vehicle stale cache hits are no longer possible.

2. Stale override state on mission switch — Fully addressed. selectionKey is wired directly into useManagedParameters and the useEffect on that key clears the override map immediately on every selection change. resetOverrides() is additionally called on category switch and on back-navigation to the mission select step. The isFirstSelectionKeyEffect guard prevents a false reset on initial render. I tested switching between Frequent Run entries and confirmed parameters update correctly each time.

3. missionPath vs composite id overloading — Fully addressed. id (composite selection key for row deduplication) and missionPath (real template path) are now explicitly separate fields on every run entry. Script loading and command creation consistently use missionPath, with a clean ?? selectedMissionId fallback that handles regular template missions without any special-casing.

Looks good to me — Approved. 

— Karen

PR #533, opened 2026-03-25
GH489 Remove Data Events from Logs by Default

ZackLyon:
- **Less noisy “Data” rows in logs** — “Data” / `dataProcessed` lines are **off by default**. Operators turn them on with **Include Data Events** in the **Filter** dropdown (not a separate control on the toolbar).
- **Filter modal** — **Include Data Events** sits **above** the **All** row; **All** checks or clears every **listed** event type. **Data** is **not** a row in the list (it’s only controlled by Include Data Events), so “All” matches what’s in the modal.
- **Default filter state** — On load, **every** modal event type is **selected**, so the log matches “show everything” without an empty checklist.
- **No types selected** — If someone clears **All** and has **no** types checked, the list is empty and a **short banner** explains opening **Filter** and picking at least one type (warning icon, shared banner styling).
- **No matches** — If filters/search leave **no** events, a second banner explains **no matching events** and suggests adjusting **Filter** or search (same visual pattern as the other banner).
- **Load more** — When the “no types selected” banner shows, **Load more** is suppressed and the virtual list count is **0** so the empty state isn’t mixed with pagination.
- **Client filtering rules** — Per-type filtering runs only when **not** “all modal types selected,” so you don’t drop events that only map to **Data** when **Data** isn’t in the checkbox list. **`dataProcessed`** is still stripped on the client when Include Data Events is off.
- **Accordion UI building block** — **`AccordionCells`** accepts **`hideBottomFade`** to hide the bottom gradient fade (for empty or message-only layouts); **optional** for other screens.
- **Tests** — **`LogFiltersDropdown`** and **`LogsSection`** tests updated for the new filter behavior, empty states, and search/no-match behavior.

### Include data events checkbox (off by default)
Screenshot 2026-03-25 at 4 36 38 PM

Note: previously the Filter modal was set up so that when nothing was selected, it would show all events by default. This made it so the behavior didn't match the checkbox values since it was showing All events, but the All checkbox wasn't selected. Now, the behavior matches the checkboxes so they are all selected by default. When the user unselects everything, it shows a warning message that no logs are being shown due to the filters.

### DIFF
--- a/apps/lrauv-dash2/components/LogsSection.tsx
+++ b/apps/lrauv-dash2/components/LogsSection.tsx
@@ -1,9 +1,5 @@
 import React, { useMemo, useState } from 'react'
-import {
-  EventType,
-  useInfiniteEvents,
-  useTethysApiContext,
-} from '@mbari/api-client'
+import { useInfiniteEvents, useTethysApiContext } from '@mbari/api-client'
 import {
   Virtualizer,
   LogCell,
@@ -18,10 +14,15 @@ import { MultiValue } from 'react-select'
 import { DateTime } from 'luxon'
 import formatEvent, {
   displayNameForEventType,
-  eventFilters,
   isUploadEvent,
 } from '../lib/formatEvent'
-import { applyEventFilters } from '../lib/eventFilterUtils'
+import {
+  applySelectedFilters,
+  defaultModalSelections,
+  deriveEventTypes,
+  hasAllNonDataFiltersSelected,
+  modalVisibleFilterIds,
+} from '../lib/logFilters'
 import { handleCopyEventLogs } from '../lib/handleCopyEventLogs'
 import { createLogger, useDebounce } from '@mbari/utils'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
@@ -51,8 +52,6 @@ const styles = {
   emptyLogBannerIcon: 'mr-1 flex-shrink-0 text-amber-600',
 }
 
-const DATA_FILTER_ID = 'Data'
-
 const LogsSection: React.FC<LogsSectionProps> = ({
   vehicleName,
   from,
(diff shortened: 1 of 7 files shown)


apps/lrauv-dash2/components/LogsSection.tsx
```diff
@@ -52,36 +65,44 @@ const LogsSection: React.FC<LogsSectionProps> = ({
...
-    () =>
-      filters.length && !allFiltersSelected
-        ? filters
-            .map(({ id }) => eventFilters[id].eventTypes)
-            .flat()
-            .filter((k, i, a) => a.indexOf(k) === i)
-        : undefined,
-    [filters, allFiltersSelected]
-  )
+  const eventTypes = useMemo((): EventType[] | undefined => {
```
jimjeffers: Codex found this and I concur with this one:

> [P1] The new Include Data Events toggle does not actually work when the user has only a subset of non-Data filters selected. In the PR branch’s apps/lrauv-dash2/components/LogsSection.tsx, eventTypes adds dataProcessed when the checkbox is on (lines 96-105), but flatData then re-filters with applyEventFilters(events, selectedFilterNames) using only the modal selections (lines 146-149). Since Data is never one of those selected names, dataProcessed events get filtered back out unless every modal filter is selected. The added tests only cover the default “all selected” case, so this regression is missed.

jimjeffers (dismissed):
We are utilizing `'dataProcessed'` and `includeDataEvents` in 4 different places and slightly differently. This needs to be refactored. Keep things DRY. The main issue is that we can get tripped up in any of the 4 places and introduce a bug here. We should just specify a module to store the filter / type defs and manage the state in the LogsSection if possible:

> **formatEvent.tsx:** Keep the eventFilters definitions there, or rename that export into a more neutral file later.
> **logFilters.ts:** Owns filter ids, default selections, and derivation/application helpers.
> **LogsSection.tsx:** Owns UI state and wiring to query/rendering. 

Log filters could define:

* type LogFilterId
* default selections
* modal-visible filter ids
* deriveEventTypes(selectedFilterIds)
* applySelectedFilters(events, selectedFilterIds)
* maybe hasAllNonDataFiltersSelected(...)

ZackLyon:
@ksalamy - Can you verify that I have addressed Jim's requested changes and approve this PR?

ksalamy (approved):
Jim requested two things in his CHANGES_REQUESTED review, and both have been fully addressed:

1. Refactor into a dedicated module — A new logFilters.ts file now owns all the filter logic (deriveEventTypes, applySelectedFilters, defaultModalSelections, etc.) exactly as Jim specified. LogsSection.tsx imports cleanly from it, and formatEvent.tsx is untouched.

2. Fix the partial-filter + Include Data toggle bug — The regression Jim identified (toggling "Include Data Events" on had no effect when only a subset of modal filters was selected) is fixed. applySelectedFilters now correctly injects the 'Data' filter name into the client-side filter pass when the toggle is on, covering all four combinations of partial/full selection and the toggle state.

Tests are well-targeted — the specific bug scenario Jim called out is now directly unit tested in logFilters.test.ts, and integration tests for the new UI states (empty state banner, no-match banner, toggle behavior) were added to LogsSection.test.tsx.

CI is passing and the PR is mergeable with no conflicts.

Approved.

PR #540, opened 2026-04-02
Fix Incorrect Vehicle Status Icon

ZackLyon:
- **Surface vs underwater:** Vehicles with **recent cell traffic** (or sat, or a live cell ping) are treated as **in contact / likely surfaced**, not only when the cell ping says reachable. That fixes cases where the UI said “likely underwater” even though cell was active but sat was stale.
- **Comms window:** Added **`isEffectivelyCommunicating`** with a configurable **recency window** so last cell and last sat timestamps count the same way as “still talking” (plus websocket ping when reachable).
- **Docked vs at sea:** Added **`getPluggedInStatus`** so “plugged in” comes from **recovery on the deployment** or **mission start still in the future** (pre-launch), matching existing deployment behavior, with unit tests.
- **Websocket → UI:** **Tethys subscription** cache keys now include the **logged-in account email** (aligned between writers and readers), and **`useTethysSubscriptionEvent`** uses **React Query** so cell ping updates **actually re-render** components instead of a one-off `getQueryData` read.
- **Single hook:** **`useVehicleStatus`** combines ping, last comms times, wall clock, and plugged-in detection, and exposes **`cellPingReachable`** (cell channel / first toolbar icon) separately from **`isLikelySurfaced`** / **`physicalStatus`** (combined “where we think the vehicle is” for the second icon and copy).
- **Deployment page:** Toolbar uses the hook for **cell connected vs not**, **vehicle status icon**, and **`VehicleInfoCell`** (`isReachable` tracks **likely surfaced**, not ping alone); **status icon** comes from **`physicalStatus`** via a small **`vehiclePhysicalStatusIcon`** helper.
- **Vehicle list:** **“Likely on surface”** next to GPS age uses the same **`isLikelySurfaced`** logic instead of raw ping reachability.
- **Tests:** Added **`effectiveVehicleComms`** and **`getPluggedInStatus`** Jest suites under `apps/lrauv-dash2/jest/`.


jimjeffers:
The new helper tests are useful, but they stop one layer short of the regression this PR is fixing. Since the change introduces a new composed status contract in `useVehicleStatus` and a new reactive path in `useTethysSubscriptionEvent`, I’d like to see at least one test at each of those boundaries.

For example, a hook-level regression that proves recent cell traffic makes the vehicle look surfaced even when the ping itself is unreachable:

```tsx
mockedUseTethysSubscriptionEvent.mockReturnValue({
  eventName: 'VehiclePingResult',
  vehicleName: 'makai',
  reachable: false,
} as any)

render(<Probe lastCellCommsTime={NOW - 60_000} />)

expect(screen.getByTestId('connected')).toHaveTextContent('false')
expect(screen.getByTestId('physical')).toHaveTextContent('surfaced')
```

And a cache-reactivity test that proves the React Query subscription actually causes a re-render when websocket data is written:

```tsx
act(() => {
  queryClient.setQueryData(
    getTethysSubscriptionQueryKey(
      'VehiclePingResult',
      'makai',
      'jim@sumocreations.com'
    ),
    { eventName: 'VehiclePingResult', vehicleName: 'makai', reachable: true }
  )
})

await waitFor(() => {
  expect(screen.getByTestId('reachable')).toHaveTextContent('true')
})
```

That would validate the user-visible behavior this PR claims to fix, not just the extracted helper predicates.

jimjeffers:
Optional refactor suggestion: I think the overall model here is good, but the status policy itself would be easier to reason about and test if it lived in one pure function, with `useVehicleStatus` staying thin around data access.

Right now the policy is split across `isEffectivelyCommunicating`, `getPluggedInStatus`, and then the precedence logic in `useVehicleStatus`.

Before:

```ts
const cellPingReachable = Boolean(pingEvent?.reachable)

const isPluggedIn = getPluggedInStatus({
  recoverEvent,
  startEventUnix,
})

const isLikelySurfaced = useMemo(
  () =>
    isEffectivelyCommunicating({
      pingReachable: cellPingReachable,
      lastSatCommsTime,
      lastCellCommsTime,
      nowMs,
    }),
  [cellPingReachable, lastSatCommsTime, lastCellCommsTime, nowMs]
)

const physicalStatus: VehiclePhysicalStatus = isPluggedIn
  ? 'pluggedIn'
  : isLikelySurfaced
  ? 'surfaced'
  : 'underwater'
```

After:

```ts
export const getVehiclePhysicalStatus = (input: {
  pingReachable?: boolean | null
  lastSatCommsTime: number | null
  lastCellCommsTime: number | null
  nowMs: number
  recoverEvent?: DeploymentEvent
  startEventUnix?: number | null
}) => {
  const cellPingReachable = Boolean(input.pingReachable)

  const isPluggedIn = Boolean(
    input.recoverEvent ||
      (input.startEventUnix != null && input.startEventUnix > input.nowMs)
  )

  const withinWindow = (t: number | null) => {
    if (t == null) return false
    const delta = input.nowMs - t
    return delta >= 0 && delta <= COMMUNICATION_RECENCY_MS
  }

  const isLikelySurfaced =
    cellPingReachable ||
    withinWindow(input.lastCellCommsTime) ||
    withinWindow(input.lastSatCommsTime)

  return {
    cellPingReachable,
    isPluggedIn,
    isLikelySurfaced,
    physicalStatus: isPluggedIn
      ? 'pluggedIn'
      : isLikelySurfaced
      ? 'surfaced'
      : 'underwater',
  }
}
```

Then the hook becomes just:

```ts
const pingEvent = useTethysSubscriptionEvent('VehiclePingResult', vehicleName ?? '')

return {
  pingEvent,
  ...getVehiclePhysicalStatus({
    pingReachable: pingEvent?.reachable,
    lastSatCommsTime,
    lastCellCommsTime,
    nowMs,
    recoverEvent,
    startEventUnix,
  }),
}
```

I think that wins because the actual policy becomes testable in one place, `nowMs` is fully injectable, and the hook stays focused on data wiring instead of also being where the business rule lives. Then I’d pair that with one regression test that simulates a cache/websocket update and proves the rendered state flips from underwater to surfaced.

One clarification on `nowMs`: I’m not suggesting changing where it comes from. In this PR, `nowMs` is already being derived at the call site via `useTick(60_000)` and passed into `useVehicleStatus(...)`. The change I’m suggesting is to use that same injected clock for all of the status policy. As written, the comms-recency path uses `nowMs`, but `getPluggedInStatus(...)` still calls `DateTime.now()` internally. Moving the policy into a pure `getVehiclePhysicalStatus(...)` helper would make both the “recent comms” and “pre-launch / plugged-in” branches use the same passed-in `nowMs`, which makes the policy deterministic and easier to test.

jimjeffers (approved):
Nothing is wrong with the implementation. But walked through the implementation with Codex and found some slight improvements you could consider making here before merging. That could make the hook easier to test.

ZackLyon:
Feedback has been addressed and merge conflicts have been resolved. Shipping.

PR #543, opened 2026-04-03
fix: Overview Map partial render and React warning on refresh

### LINKED ISSUE #548: fix: Overview Map renders at partial width after page refresh
## Problem

After a hard page refresh on the Overview page, the Leaflet map renders at approximately 25% of its intended width rather than filling its container. A second navigation or manual browser resize would correct it.

Additionally, the following React warning appeared in the browser console:

> Can't perform a React state update on a component that hasn't mounted yet

Both issues were caused by `[200, 800]ms` `setTimeout` calls in the `onMapReady` handler that fired `map.invalidateSize()` asynchronously — racing against component teardown and firing before the container had stabilized.

## Steps to reproduce

1. Open the Overview page (`/`)
2. Hard-refresh (Cmd+Shift+R / Ctrl+Shift+R)
3. Observe the map container renders at partial width
4. Open the browser console — observe the React state update warning

## Expected behavior

The map fills its full container width immediately on load and on refresh, with no React warnings in the console.

## Fix

Addressed in PR #543 — replaced the `setTimeout` approach with a `useResizeObserver` hook so `map.invalidateSize()` fires at the correct time and is cleaned up on unmount.

ksalamy:
Fixes #548

## Summary

- Replaces the unsafe `[200, 800]ms setTimeout` calls in `onMapReady` with a proper `useResizeObserver` hook so `map.invalidateSize()` fires at the right time and is cleaned up on unmount.
- Fixes the **map rendering at ~25% width** after a page refresh (the old timeouts were racing against component teardown).
- Eliminates the **React warning** `Can't perform a React state update on a component that hasn't mounted yet` — the async setTimeout callbacks in the render prop were the cause.

## What changed

`apps/lrauv-dash2/pages/index.tsx`
- Added `mapContainerRef` (`useRef<HTMLDivElement>`) attached to the wrapping `div` around `<Map>`.
- Replaced the manual `ResizeObserver` + `useEffect` setup with `useResizeObserver` from `@mbari/utils` (already polyfilled and throttled at 100ms); a small `useEffect` watches `size` and calls `mapRef.current.invalidateSize()` whenever the container resizes.
- `onMapReady` now stores the map instance in `mapRef` and calls `map.invalidateSize()` once synchronously (covers the case where the map becomes ready after the initial resize event).
- Removed the `[200, 800]ms` setTimeout calls.

## Test plan

- [ ] Hard-refresh the Overview page — map should fill its full container immediately
- [ ] Check browser console — the "Can't perform a React state update" warning should be gone
- [ ] Navigate away and back — map should still render at full size
- [ ] Verify center-on buttons (stations, polygons) still work on the Overview map

### DIFF
--- a/apps/lrauv-dash2/components/MapFlyTo.tsx
+++ b/apps/lrauv-dash2/components/MapFlyTo.tsx
@@ -1,10 +1,10 @@
 import React, { useEffect } from 'react'
 import { useMap } from 'react-leaflet'
-import { useSelectedStations } from './SelectedStationContext'
+import { useMapCamera } from './MapCameraContext'
 
 const MapFlyTo: React.FC = () => {
   const map = useMap()
-  const { flyToRequest, setFlyToRequest } = useSelectedStations()
+  const { flyToRequest, setFlyToRequest } = useMapCamera()
 
   useEffect(() => {
     if (flyToRequest) {
--- a/apps/lrauv-dash2/pages/index.tsx
+++ b/apps/lrauv-dash2/pages/index.tsx
@@ -2,7 +2,6 @@ import { OverviewToolbar } from '@mbari/react-ui'
 import { NextPage } from 'next'
 import Layout from '../components/Layout'
 import React, { useCallback, useEffect, useRef, useState } from 'react'
-import ResizeObserver from 'resize-observer-polyfill'
 import dynamic from 'next/dynamic'
 import VehicleDeploymentDropdown from '../components/VehicleDeploymentDropdown'
 import VehicleList from '../components/VehicleList'
@@ -22,7 +21,7 @@ import { MapLayersListModal } from '../components/MapLayersListModal'
 import { useSelectedStations } from '../components/SelectedStationContext'
 import { useMarkers } from '../components/MarkerContext'
 import toast from 'react-hot-toast'
-import { createLogger } from '@mbari/utils'
+import { createLogger, useResizeObserver } from '@mbari/utils'
 import { PlatformsListModal } from '../components/PlatformsListModal'
 import { useRefreshPositions } from '../lib/useRefreshPositions'
 import VehicleColorsModal from '../components/VehicleColorsModal'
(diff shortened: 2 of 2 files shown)


ksalamy:
Hi @ZackLyon  — this is ready for your review. 

The code went through 3 rounds of Copilot feedback, all resolved:

- Initial pass flagged the requestAnimationFrame in onMapReady as a potential post-unmount race → simplified to a synchronous map.invalidateSize() call.
- Copilot then flagged the raw ResizeObserver usage as unthrottled and relying on a global → replaced the whole manual setup with useResizeObserver from @mbari/utils, which is already polyfilled and throttled (100 ms).
- Replaced the try/catch guard (which was silently swallowing unrelated errors) with a simple if (!mapRef.current) return.

Net result: ~20 lines removed, the @mbari/utils shared hook does the heavy lifting, and Copilot gave a clean pass. 

The branch is rebased on current develop with no conflicts.

Thanks!
Karen

ZackLyon (approved):
@ksalamy - This is approved but please review the following process notes for future PRs:

1) It looks like this PR builds on work from #537, but that work wasn’t included in this branch. I had to pull the additional changes from develop to get things building locally.

In situations like this, it’s usually best to either:
	•	Wait for the original PR (#537) to merge into develop, then create a new branch, or
	•	Create a stacked PR by branching this work directly off the #537 branch

This helps ensure the PR builds cleanly and makes the dependency between changes clearer.

2) For future fixes, please open an issue before submitting a PR. The summary here helped clarify the problem, but having an issue first provides clearer context and documents the problem.

ksalamy:
Hi @ZackLyon,

Thanks so much for your thoughtful review and approval despite a few "hiccups" on my end. Much appreciated.  

**Both of the process concerns flagged in the review have been addressed:**

**Build dependency on #537 — resolved. #537 merged into develop on April 8, so this branch now builds cleanly against the current develop baseline without requiring any manual cherry-picks.

Missing issue — resolved. Issue #548 ("fix: Overview Map renders at partial width after page refresh") has been opened and linked in the PR description (Fixes #548). It will close automatically on merge.**

The fix itself is unchanged from the original submission — onMapReady's unsafe setTimeout-based invalidateSize() calls replaced with a useResizeObserver hook, eliminating both the partial-width render on refresh and the React unmounted-component warning. No other files were modified.

PR #547, opened 2026-04-08
GH262 Docs Section Defaults to This Vehicle

ZackLyon:
- Make "This Vehicle" the default for the DocsSection dropdown
- Update tests to match new default selection

### Default Docs Section view for Sim
Screenshot 2026-04-08 at 4 22 40 PM

### DIFF
--- a/apps/lrauv-dash2/components/DocsSection.test.tsx
+++ b/apps/lrauv-dash2/components/DocsSection.test.tsx
@@ -61,7 +61,7 @@ describe('DocsSection', () => {
     expect(() =>
       render(
         <MockProviders queryClient={new QueryClient()}>
-          <DocsSection />
+          <DocsSection vehicleName="pontus" />
         </MockProviders>
       )
     ).not.toThrow()
@@ -70,7 +70,7 @@ describe('DocsSection', () => {
   test('should render the name of one of the documents', async () => {
     render(
       <MockProviders queryClient={new QueryClient()}>
-        <DocsSection />
+        <DocsSection vehicleName="pontus" />
       </MockProviders>
     )
     await waitFor(() => {
@@ -82,7 +82,7 @@ describe('DocsSection', () => {
   test('should not render the add note button if not authenticated', async () => {
     render(
       <MockProviders queryClient={new QueryClient()}>
-        <DocsSection />
+        <DocsSection vehicleName="pontus" />
       </MockProviders>
     )
     expect(screen.queryByText(/add document/i)).not.toBeInTheDocument()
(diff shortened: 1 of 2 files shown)


ZackLyon:
@ksalamy - Would you mind reviewing this PR? Jim has been pretty busy.

jimjeffers:
Hey Zack, a minor optimization: you could make this state change in the parent component instead of hardcoding the default inside `DocsSection`.

Right now the PR changes the shared components initial state to `"This Vehicle"`. That works for the current callers, but it makes the default implicit for every future use of `DocsSection`.

A cleaner alternative would be to keep `DocsSection` generic and pass something like `initialFilterType="This Vehicle"` from the vehicle-specific parents, such as `VehicleAccordion` and `DocsModal`. That preserves the exact UX this issue wants, while keeping the component flexible if we ever reuse it in an overview/global docs context.

For example, instead of this inside `DocsSection`:

```tsx
const [selectedType, setSelectedType] =
  useState<DocumentFilterType>("This Vehicle")
```

You could make the component configurable:

```tsx
interface DocsSectionProps {
  authenticated?: boolean
  vehicleName: string
  currentDeploymentId?: number
  initialFilterType?: DocumentFilterType
}

const DocsSection: React.FC<DocsSectionProps> = ({
  authenticated,
  vehicleName,
  currentDeploymentId,
  initialFilterType = "All Documents",
}) => {
  const [selectedType, setSelectedType] =
    useState<DocumentFilterType>(initialFilterType)
```

And then opt into the behavior explicitly from the vehicle-specific parents:

```tsx
<DocsSection
  authenticated={authenticated}
  vehicleName={vehicleName}
  currentDeploymentId={currentDeploymentId}
  initialFilterType="This Vehicle"
/>
```

```tsx
<DocsSection
  vehicleName={vehicleName}
  initialFilterType="This Vehicle"
/>
```

It would also make the tests more intent-driven: assert that the selected filter is `"This Vehicle"` and that non-matching vehicle docs are hidden, instead of only updating render inputs so the existing assertions still pass.

jimjeffers (approved):
Approved. I walked through this with Codex this morning, and it looks good. There's a minor API optimization we could make that would make the component more future-proof by allowing the calling component to specify the initial filter type, since it knows what it actually wants. We only have one use case, which is why it's not apparent, but if that changes, you'll need to pull this out in the future. So, the codex is right, that's a better design.

ZackLyon:
In general I agree with that feedback, but in this instance I don't think the extra complexity is warranted yet since there are no plans to re-use this component at this time. It will be minor change in the future if this component is ever re-used. Also, I can't remember which local repo this branch is on.

Issue #545, opened 2026-04-07
Enhancement: Add Quick-Access Resources dropdown near PIC section


ksalamy:
## Summary

Add a dropdown menu near the PIC section of the dashboard that gives pilots and operators fast, in-context access to operational documents, training resources, and (for admins) frontend configuration settings — without leaving dash5.

## Motivation

Currently, users must navigate away from dash5 or search externally for documents such as the LRAUV Watchbill signup, PIC training refreshers, and other operational references. Keeping these resources one click away reduces context switching during active operations.

### 1. Quick-Access Resources Dropdown
- Positioned adjacent to the PIC section in the UI
- Contains links/shortcuts to operational documents, including:
  - LRAUV Watchbill signup list
  - Any other mission-critical or reference documents (configurable)

### 2. PIC Training / Formative Links
- A subsection (or grouped items) linking to the Formative website for PIC training refreshers
- Could include links to other pilot certification or procedure resources

### 3. Admin-Only Frontend Settings Subsection
- A protected subsection visible only to admin-role users
- Allows frontend configuration changes (e.g. feature flags, display preferences, defaults) without a code deploy
- Should be gated behind role/permission check so non-admins do not see it

## Open Questions

- Where exactly should the dropdown anchor — inline with the PIC label, or as a separate icon/button nearby?
- Should document links be hardcoded, pulled from a config file, or managed dynamically via API?
- What is the source of truth for admin role detection (existing auth context, environment variable, etc.)?
- Should the Formative links be per-vehicle/per-pilot or global?

## Acceptance Criteria

- [ ] Dropdown renders next to (or near) the PIC section
- [ ] At minimum, Watchbill signup and Formative training links are accessible from the dropdown
- [ ] Admin-only subsection is hidden for non-admin users and visible for admin users
- [ ] Links open in a new tab (do not navigate away from dash5)
- [ ] Component is accessible (keyboard navigable, ARIA labels)

## Labels
`enhancement`, `ux`, `admin`

ksalamy:
Resolved in develop — Quick-Access Resources dropdown implemented and merged via PR #549 (refactored per Zack's review in #54e8c9c).

PR #549, opened 2026-04-09
feat(#545): Add Quick-Access Resources dropdown near PIC section

ksalamy:
## Summary

Implements [issue #545](https://github.com/mbari-org/dash5/issues/545) — a Quick-Access Resources dropdown placed near the PIC section in the `OverviewToolbar`.

- Only visible to authenticated users; Admin Settings section only shown to users with the `admin` role
- Folder icon trigger toggles between closed/open states with an "LRAUV Resources" tooltip
- Sections are collapsible — PIC Signup is open by default, all others start collapsed
- All links open in a new tab with per-link icons and tooltips
- Full LRAUV link set wired into `lrauv-dash2`: Watchbill, Quick Start, Operator Reference, Cheat Sheet, Vehicle Config Worksheet, TethysDash Operation, Training Worksheet, and Operator Refresher Quiz

**Also addressed during review:** `aria-controls` validity, stable React keys, `useCallback` for click-outside handler, static link arrays hoisted to module scope, `useId()` for panel IDs, `type="button"` on `IconButton`, `ResourceLink` union type, `zIndex` Tailwind config placement, `top-full` panel positioning, and 24-hour time test alignment.

## Test plan

- [ ] Confirm dropdown is **not visible** when logged out
- [ ] Confirm dropdown **appears** after login
- [ ] Confirm sections collapse/expand on header click; PIC Signup is open by default
- [ ] Confirm Admin Settings section is **hidden** for non-admin users and **visible** for admin users
- [ ] Confirm all links open in a new tab
- [ ] Confirm folder icon toggles open/closed on click
- [ ] Confirm "LRAUV Resources" tooltip appears above the icon


ksalamy:
Hi @ZackLyon,

PR #549 is ready for your review when you get a chance: https://github.com/mbari-org/dash5/pull/549

This adds a Quick-Access Resources dropdown near the PIC section (issue #545). 

**This PR is an enhancement to the Dash5 UI:**

- A folder icon appears in the OverviewToolbar only when the user is authenticated
- Clicking it opens a dropdown with collapsible sections — LRAUV PIC Signup (open by default), Resources, Operator Training, and an Admin Settings section visible only to admins
- Links include Watchbill, Operator Reference, Quick Start, Cheat Sheet, Vehicle Config Worksheet, TethysDash Operation, Training Worksheet, and Operator Refresher Quiz
- All links open in a new tab
- The PR went through several rounds of Copilot review and all comments have been addressed. Tests are green.

Would love your feedback — particularly on the component structure in ResourcesDropdown.tsx and how it's wired into [...deployment].tsx.

Thanks! Karen

ZackLyon (approved):
This is a solidly working implementation, but the architectural issue I see is with how `ResourcesDropdown` is being treated as a reusable component while it is structured like a one-off component, so it isn't really benefiting from the advantages of either. 

As a reusable component, it doesn't iterate over objects in an array to create the `CollapsibleSection` components, so it is only able to be used in this one situation where there are these 4 specific sections yet still relies heavily on prop drilling to receive labels, urls, etc. 

As a one-off specific component, it adds a huge section of tightly coupled code to [...deployment].tsx instead of owning that code itself.

There are a couple ways to resolve this and trade-offs to each:
1) The simplest/least amount of extra code would be to move `ResourcesDropdown` into  the `lrauv-dash2/components` directory, then include the urls/labels in the component. That component could then be passed as a `React.Node` prop into `OverviewToolbar` and would look like `<ResourcesDropdown isAdmin={isAdmin}/>` as a prop. This acknowledges it as a one off component that isn't tightly coupled to other components.
2) Refactor `ResourcesDropdown` to accept an array of `CollapsibleSection` props and make it reusable, then make a wrapper component that essentially follows all the advice from 1.
3) This is a bigger refactor, but might be worth it if you think more things will be added to `OverviewToolbar` in the future. Move `OverviewToolbar` into `lrauv-dash2` where it can use API hooks instead prop drilling from the parent component (ie [...deployment].tsx in this case). In my opinion, this is the best move architecturally with the biggest payoff for future development, but it is also the largest refactor.

This is a non-blocking review, but I would highly consider either doing one of the above items or creating a separate issue to follow up. FYI: These are my own thoughts and Jim or AI might have different recommendations.

PR #552, opened 2026-04-13
GH289 Argos Battery Indicator

ZackLyon:
Add Argos battery indicator to vehicle diagrams

**IMPORTANT NOTE**: Brian's comment on GH#289 indicated that that there is a `text_argoago` string that is returned from the API, but I was unable to locate this in either the dash4 code or the auv repo code. I do not see any argo text indicators on the dash4 vehicle diagrams, so I assume the battery indicator should be sufficient.

### Vehicle diagrams with argo battery indicator
Screenshot 2026-04-13 at 3 22 31 PM

### DIFF
--- a/apps/lrauv-dash2/components/VehicleDiagram.tsx
+++ b/apps/lrauv-dash2/components/VehicleDiagram.tsx
@@ -120,6 +120,7 @@ const VehicleDiagram: React.FC<{
         colorAmps={vehicle?.color_amps}
         colorDvl={vehicle?.color_dvl}
         textGpsAgo={vehicle?.text_gpsago}
+        colorArgo={vehicle?.color_argo}
         textCellAgo={formattedCellAgo}
         textNoteTime={vehicle?.text_notetime}
         textArrow={vehicle?.text_arrow}
--- a/apps/lrauv-dash2/components/VehicleList.tsx
+++ b/apps/lrauv-dash2/components/VehicleList.tsx
@@ -203,6 +203,7 @@ const ConnectedVehicleCellComponent: React.FC<{
         colorAmps: vehicle.color_amps,
         colorDvl: vehicle.color_dvl,
         textGpsAgo: vehicle.text_gpsago,
+        colorArgo: vehicle.color_argo,
         textCellAgo: formattedCellAgo,
         textNoteTime: vehicle.text_notetime,
         textArrow: vehicle.text_arrow,
--- a/apps/lrauv-dash2/styles/vehicle.css
+++ b/apps/lrauv-dash2/styles/vehicle.css
@@ -143,17 +143,17 @@
     fill: #333333; 
   }
 
-	.st32 {
-    fill: #aaaaaa;
-    stroke: #000000;
-  }
-
 	.st30 {
     font-size: 8px;
   }
 
 	.st31 {
     fill: #B4372D;
+  }
+
+	.st32 {
+    fill: #aaaaaa;
+    stroke: #000000;
   }
 
 	.stleak2 {
--- a/apps/lrauv-dash2/styles/vehicle.css
+++ b/apps/lrauv-dash2/styles/vehicle.css
@@ -143,6 +143,11 @@
     fill: #333333; 
   }
 
+	.st32 {
+    fill: #aaaaaa;
+    stroke: #000000;
+  }
+
 	.st30 {
     font-size: 8px;
   }
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -2,7 +2,7 @@
   "name": "@mbari/api-client",
   "version": "1.0.0",
   "source": "src/index.ts",
-  "types": "./dist/index.d.ts",
+  "types": "./src/index.ts",
   "files": [
     "dist"
   ],
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -2,7 +2,7 @@
   "name": "@mbari/api-client",
   "version": "1.0.0",
   "source": "src/index.ts",
-  "types": "./src/index.ts",
+  "types": "./dist/index.d.ts",
   "files": [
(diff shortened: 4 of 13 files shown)


ZackLyon:
@ksalamy - Can you please review when you have time. Thanks!

ksalamy (commented):
## Code Review — PR #552 Argos Battery Indicator

**Reviewer:** @ksalamy
**Date:** April 17, 2026
**Tested locally:** Yes — branch checked out, built, and run against live okeanids.mbari.org data

---

### Overview

This is a well-scoped, focused PR. The `ArgosBatteryIndicator` component is well-built and the tests all pass cleanly (108/108). Two issues need to be addressed before merge — one blocking, one minor build issue — plus a few small polish items.

---

## BLOCKING — Must fix before merge

### 1. Argos indicator is gray in the deployment map view but green in the vehicle overview — core feature not working correctly

Confirmed with live data on Ahi and Daphne: both vehicles show a green Argos battery in the vehicle overview cards, but the indicator is grayed out in the full-width deployment map diagram. This is a direct regression from dash4, which shows the correct battery color in all views.

**Root cause — `isDocked` guard in `ArgosBatteryIndicator` incorrectly suppresses the color:**

```tsx
// Current — wrong
const innerClass = isDocked ? 'st18' : colorArgo ?? 'st18'
```

When `isDocked` is `true`, the API's `color_argo` value is completely ignored and the fill is forced to `st18` (no stroke, no fill — invisible). The API confirms it returns `color_argo: "st25"` for Daphne even in the plugged-in state, meaning the server intends the indicator to be shown regardless of dock state. Dash4 shows it. The fix is a one-liner:

```tsx
// Fix — always respect what the server sends
const innerClass = colorArgo ?? 'st18'
```

Once this guard is removed, `isDocked` serves no purpose in this component and the `isDocked` prop can be removed from `ArgosBatteryIndicatorProps` entirely.

**Why the two views disagree:**

The overview (`VehicleList.tsx`) derives `status` from deployment event data — for Daphne, `lastDeployment?.recoverEvent` is null at render time, so `status = 'onMission'`, `isDocked = false`, and Argos shows green.

The deployment view (`VehicleDiagram.tsx`) derives `status` from the live mission text string — Daphne's `text_mission` contains `"PLUGGED"`, so `status = 'pluggedIn'`, `isDocked = true`, and Argos is suppressed.

The inconsistent `status` computation between the two views is pre-existing, but the `isDocked` guard is what turns it into a visible inconsistency. Removing the guard fixes both views.

---

### 2. PR breaks a clean build — dist files are stale

Anyone pulling this branch and running `yarn workspace @mbari/lrauv-dash2 build` without first manually rebuilding dependencies will see:

```
Type error: Property 'colorArgo' does not exist on type 'IntrinsicAttributes & FullWidthVehicleDiagramProps'.
Type error: Property 'color_argo' does not exist on type 'GetVehicleInfoResponse'.
```

Both `packages/react-ui` and `packages/api-client` resolve their types from `dist/index.d.ts` per each package's `"types"` field in `package.json`. The PR adds `colorArgo` and `color_argo` to both packages' source but does not update the compiled `dist/`. CI and any colleague checking out the branch will hit this immediately. Either commit the rebuilt `dist/` files for both packages, or remove the `"types": "./dist/index.d.ts"` fields so Next.js always resolves directly from `src/index.ts`.

---

## MINOR

### 3. `vehicle.css` — `st32` inserted out of numerical order

`st32` is placed before `st30`. Small thing — move it after `st30` to keep the file sorted.

### 4. Storybook stories don't cover all Argos states

Only the `Deployed` story has `colorArgo: 'st27'`. Consider adding `colorArgo: 'st25'` to `Scheduled` to show the healthy state. Once the `isDocked` guard is removed, also verify that `Recovered` and `PluggedIn` show the correct color when the API provides one.

### 5. `Vehicle.test.tsx` only tests `st27`, not `st25`

`FullWidthVehicleDiagram.test.tsx` tests both healthy (`st25`) and low (`st27`) states. `Vehicle.test.tsx` only covers `st27`. A test for `colorArgo='st25'` confirming the `'Argos battery OK'` aria label would round it out.

---

## What's good

The `ArgosBatteryIndicator` component is well-built. `role="alert"` on `st27` is a nice a11y touch, `aria-hidden` when decorative is correct, and the aria label logic distinguishing OK / low / unknown states is solid. The JSDoc comment referencing the Python SVG source is helpful. `color_argo?: string` being optional in `GetVehicleInfoResponse` is the right defensive typing. All 108 tests pass.

---

## Summary

| Severity | Count |
|---|---|
| Blocking | 2 |
| Minor | 3 |

One-line fix to `ArgosBatteryIndicator` + updating the dist files and this is ready to go.

ksalamy:
Hey @ZackLyon ,

Really nice work on the ArgosBatteryIndicator component — the accessibility touches (role="alert", aria-hidden, the aria label states) are exactly right and all 108 tests pass. Well scoped PR too.

**Two things to fix before we can merge:**

The main one: the indicator shows green in the vehicle overview cards but grays out in the full-width deployment map view for both Daphne and Ahi. One-line fix in ArgosBatteryIndicator — remove the isDocked ? 'st18' guard so the component always respects what the server sends. The API returns color_argo: "st25" for Daphne even when she's plugged in, so the server clearly intends it shown in all states, just like dash4.

The build: the dist/ type files for react-ui and api-client are stale — TypeScript can't find colorArgo or color_argo on a clean pull. CI will hit this too.

A few minor polish items in the review comment as well (CSS ordering, Storybook coverage, one missing test case).

Fix those two things and this is good to go! 

Karen

ksalamy (changes requested):

ksalamy (approved):
All review items addressed — isDocked guard removed, types resolve from source, CSS ordering fixed, Storybook and test coverage added. 487 tests passing. Good to merge.

PR #561, opened 2026-04-20
fix(api-client): wrap matchAll in Array.from to fix es5 target iteration error

ksalamy:
## Summary

- `String.prototype.matchAll` returns a `RegExpStringIterator` which cannot be used in `for...of` when tsconfig targets `es5` without `downlevelIteration`
- Wrapping in `Array.from()` resolves the iterator to a plain array before iterating — no config changes needed
- All 13 `extractOverrides` tests pass

## Root cause

`apps/lrauv-dash2/tsconfig.json` extends `tsconfig/nextjs.json` which sets `"target": "es5"`. Next.js 15 type-checks across workspace package source files, surfacing this pre-existing issue in `extractOverrides.ts`.

### DIFF
--- a/packages/api-client/src/axios/Util/extractOverrides.ts
+++ b/packages/api-client/src/axios/Util/extractOverrides.ts
@@ -80,7 +80,7 @@ export const extractOverrides = (
   const regex = /(?:set\s+)?([\w]+)[.:]([\w.]+)\s+([^;"']+)/g
   const commands: Override[] = []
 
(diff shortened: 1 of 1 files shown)


ksalamy:
Pre-existing String.prototype.matchAll iterator incompatibility with tsconfig target: es5 — surfaced when Next.js 15 began type-checking across workspace package sources. One-line fix wrapping matchAll in Array.from(). All 13 extractOverrides tests pass, CI green. Merging to unblock staging deploy.

PR #578, opened 2026-05-01
fix: restore California Offshore Platform icons on map

ksalamy:
## Summary

- **Widen platform position query window** from 24h to 365 days so fixed/infrequently-updated platforms (e.g. CA offshore structures) are found even when their last fix is months old
- **Pass `iconUrl` through to the map** — `GetPlatformsResponse.iconUrl` was being dropped in `PlatformPaths`; now passed to `PlatformPath` with the correct `https://odss.mbari.org/odss/` prefix (matching `PlatformSection`)
- **Use `divIcon` with `onerror`** instead of `L.icon()` to prevent broken-image placeholders when ODSS blocks direct browser image requests via CORS
- **Constrain platform icon to 32×32** to prevent mooring icons from rendering at their full natural size
- **Guard `response?.results` in elevation service** against `undefined` when Google Maps is not initialized
- **Restore direct TethysDash fetch** for Google Maps API key as a fallback (needed when TethysApi context is unavailable, e.g. local dev); swallowed silently so Next.js dev-mode overlay no longer shows it as an unhandled error
- **Fix Leaflet default marker icon 404s** — copy `marker-icon.png`, `marker-icon-2x.png`, `marker-shadow.png` to `/public` and configure `L.Icon.Default` to point at them (webpack drops Leaflet's CSS image references during bundling)

## Test plan

- [ ] Open the OverviewMap and click **TrackDB** → select California Offshore Platforms → click Apply
- [ ] Verify platform icons render on the map without broken-image placeholders
- [ ] Verify mooring icons are not oversized (should be ~32px)
- [ ] Verify the Google Maps satellite layer loads correctly
- [ ] Verify no `marker-icon-2x.png` or `marker-shadow.png` 404s in the console
- [ ] Verify no `TypeError: Failed to fetch` unhandled runtime error on page load
- [ ] Verify no `TypeError: Cannot read properties of undefined (reading 'results')` from elevation service


ksalamy:
Updated PR description to clarify that `DEFAULT_LOOKBACK_DAYS = 30` (not 365 as previously stated). The 30-day window balances coverage for infrequently-updated fixed platforms without excessive TrackDB load.

PR #645, opened 2026-05-18
feat(#217): show deployment start date below deployment title in toolbar

### LINKED ISSUE #217: Show small date below deployment info
Add time associated with deployment (e.g  _1 day ago_ ) next to the deployment info? Would help know if this is an ongoing or older event.

(Added in screenshot.)

DesktopModified

ksalamy:
## Summary

Closes #217

Adds a small **"Started X ago"** / **"Starts in X"** subtitle permanently visible beneath the deployment name in `OverviewToolbar`, so users can immediately see whether a deployment is recent or historical without having to open the deployment dropdown.

- Subtitle is always visible (not only inside the dropdown) using Luxon's `toRelative()` for human-friendly phrasing.
- Removed the now-redundant date from the dropdown header (the dropdown still shows the deployment name as a header for orientation when browsing).
- Added `aria-label="deployment date"` on the subtitle span for accessibility.
- Updated `OverviewToolbar.test.tsx`: replaced tests that only checked for the date inside the dropdown with tests that verify the date is visible at all times; added a dedicated regression test.

## Test plan

- [x] `yarn workspace @mbari/react-ui test` — all 24 `OverviewToolbar` tests pass
- [x] Dev server confirms the subtitle renders on the deployment page toolbar (visible on page load, no interaction required)

### DIFF
--- a/packages/react-ui/src/Toolbars/OverviewToolbar.test.tsx
+++ b/packages/react-ui/src/Toolbars/OverviewToolbar.test.tsx
@@ -93,7 +93,7 @@ test('should show the deployment start date below the title without opening the 
   render(<OverviewToolbar {...props} />)
 
   // The subtitle is always visible — no click needed
-  expect(screen.getByLabelText('deployment date')).toBeInTheDocument()
+  expect(screen.getByTestId('deployment-date')).toBeInTheDocument()
   expect(screen.getByText(/3 days ago/i)).toBeInTheDocument()
 })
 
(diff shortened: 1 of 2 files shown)


ksalamy:
The `aria-label="deployment date"` was intentionally removed in the last commit: a static `aria-label` that differs from the visible text overrides the accessible name and can confuse screen readers. The visible text ('Started 3 days ago') already serves as the accessible name without any override. `data-testid` is used solely for test targeting and has no accessibility effect. The PR description referenced the original approach before that fix — no code change needed.

ksalamy:
As noted in my earlier comment, this is intentional. The PR description was written before we removed `aria-label` per Copilot's own prior feedback (the static label was masking the visible text from screen readers). The visible text IS the accessible name — no `aria-label` is needed or wanted. Requesting re-review with no code changes.

PR #648, opened 2026-05-19
fix: hide operator attribution for automatic Default missions in Schedule History

ksalamy:
## Summary

The Tethys API stamps the authenticated HTTP session user on all events — including automatic Default missions triggered by the vehicle's own MissionManager. This caused Schedule History to display the browsing operator's name as **Operator** for a mission they never sent.

**What is fixed:** The synthetic Default mission rows injected by \`ScheduleSection\` from \`missionTimeline\` data already carry \`isDefaultMission: true\`. The modal's dedicated Default-mission rendering path (early return) has never shown an Operator field. The list row (\`ScheduleCell\`) previously showed the session user as \`name\` — this PR clears that to an empty string so no attribution appears.

**Known limitation:** API \`run\` events for Default (e.g. \`load Default.tl;run\` stamped with the session user by Tethys) cannot be safely distinguished from operator-intentional Default sends without backend metadata. Those events are left unchanged.

## Changes

- \`ScheduleSection\`: pass \`name=""\` on the Default mission \`ScheduleCell\` list row (synthetic path only)
- \`ScheduleCell\` (react-ui): conditionally render the name \`<li>\` only when non-empty — prevents a blank line
- Tests: synthetic Default row hides Operator in modal; \`ScheduleCell\` empty-name produces no DOM element

## Test plan
- [ ] Schedule History — synthetic Default mission row shows no operator name in the list
- [ ] Click the synthetic Default mission row → details modal shows no Operator field
- [ ] Any operator-sent mission → Operator field still shows the sender

### DIFF
--- a/apps/lrauv-dash2/components/ScheduleEventDetailsModal.tsx
+++ b/apps/lrauv-dash2/components/ScheduleEventDetailsModal.tsx
@@ -809,15 +809,21 @@ export const ScheduleEventDetailsModal: React.FC<
             </p>
             <p className="font-medium">{formatVia(event.via)}</p>
           </div>
-          {!event.isDefaultMission &&
-            event.label?.trim().toLowerCase() !== 'default mission' && (
+          {(() => {
+            const lbl = event.label?.trim().toLowerCase() ?? ''
+            const isDefault =
+              event.isDefaultMission ||
+              lbl === 'default mission' ||
+              lbl === 'default'
+            return isDefault ? null : (
               <div>
                 <p className="text-sm uppercase tracking-wide text-stone-500">
                   Operator
                 </p>
                 <p className="font-medium">{event.user || 'Unknown'}</p>
               </div>
-            )}
+            )
+          })()}
           <div>
             <p className="text-sm uppercase tracking-wide text-stone-500">
               Event ID
--- a/apps/lrauv-dash2/components/ScheduleEventDetailsModal.tsx
+++ b/apps/lrauv-dash2/components/ScheduleEventDetailsModal.tsx
@@ -809,14 +809,12 @@ export const ScheduleEventDetailsModal: React.FC<
             </p>
             <p className="font-medium">{formatVia(event.via)}</p>
           </div>
-          {!event.isDefaultMission && (
-            <div>
-              <p className="text-sm uppercase tracking-wide text-stone-500">
-                Operator
-              </p>
-              <p className="font-medium">{event.user || 'Unknown'}</p>
-            </div>
-          )}
+          <div>
+            <p className="text-sm uppercase tracking-wide text-stone-500">
+              Operator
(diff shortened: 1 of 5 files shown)


ksalamy:
Both comments addressed in subsequent commits:

1. **Operator field hidden vs. 'Automatic (vehicle)'** — after discussion with the team, the decision was to omit the Operator row entirely for Default missions rather than show a placeholder. Default missions already have their own modal path (the `isDefaultMission` branch) which never showed an Operator row; this change brings the label-based API run event path into parity with that. Showing 'Automatic (vehicle)' would add noise to a row that is already clearly described as 'Automatic fallback'.

2. **Missing tests** — three regression tests were added covering: `isDefaultMission: true` hides the row, `label: 'Default Mission'` with `isDefaultMission: false` hides the row (the exact API run event path that surfaced the original bug), and a normal operator-sent mission still shows the Operator field. All 11 tests pass.

ksalamy:
Copilot's comments are based on the original PR title/description which said 'Automatic (vehicle)'. The implementation was intentionally changed after discussion — **the Operator row is omitted entirely** rather than replaced with a placeholder.

Rationale: Default missions are already clearly identified by the 'Default Mission / Automatic fallback' label on the row and in the modal. Adding any Operator value (real or synthetic) misleads operators into thinking someone commanded it. The row simply does not apply to automatic vehicle behavior, so it is not shown.

All three tests correctly assert the agreed-upon behavior:
- Operator row hidden when `isDefaultMission: true`
- Operator row hidden when `label: 'Default Mission'` (the API run-event path that surfaced the original bug)
- Operator row still visible for regular operator-sent missions

The PR description has been updated to reflect this decision.

ksalamy:
All three Copilot comments are now addressed:

1. **'Automatic (vehicle)' vs hiding** — the PR title has been updated to reflect the actual design decision: the Operator row is omitted entirely for Default missions, not replaced with a placeholder. This was an intentional product decision made after the original commit.

2. **Tests asserting wrong behavior** — the tests correctly assert the agreed behavior (Operator hidden). The PR description now matches.

3. **Blank `<li>` from `name=""`** — fixed in `ScheduleCell.tsx`: the name `<li>` is now conditionally rendered only when `name` is non-empty, so passing an empty string produces no DOM element.

ksalamy:
The PR description has been updated to accurately reflect the implemented behavior.

The modal only branches on `event.isDefaultMission` — no label matching. The test correctly covers only that path. A label-only `'Default Mission'` event without the flag would still show Operator, and that is intentional: label-based detection was reverted after several rounds of review because it could incorrectly suppress attribution for operator-sent missions with that name. The only safe suppression is the explicit `isDefaultMission: true` flag, which is set exclusively by the synthetic injection path in `ScheduleSection`.

PR #658, opened 2026-06-01
feat(#652): dim historical track and highlight future path during timeline scrubbing

ksalamy:
## Summary

Closes #652

- When scrubbing the mission progress timeline, the vehicle GPS track now splits at the scrub position:
  - **Past segment** (already traveled): solid line at 35% opacity — de-emphasized to show history
  - **Future segment** (still to navigate): bright dashed line — highlighted to show what lies ahead
- Also fixes an unhandled `MapsServerError` from the Google Elevation API (`ELEVATION_LOCATIONS: UNKNOWN_ERROR`) that was crashing the Next.js dev overlay; added `.catch()` guards in `MouseCoordinates` and `MapDepthDisplay` so transient Google Maps errors are silently absorbed upstream where they are already logged/handled.

## Test plan

- [ ] Open a vehicle deployment page with an active GPS track
- [ ] Hover the mission progress bar at the bottom — track should split: solid/dim past, dashed/bright future
- [ ] Move to left end of timeline — most of track should be dashed/bright (future)
- [ ] Move to right end — most of track should be solid/dim (past)
- [ ] Move mouse off timeline — full solid track restores after ~1 second
- [ ] Hover the deployment map — confirm no `MapsServerError` runtime error overlay appears

Issue #652, opened 2026-06-01
Timeline scrubbing should visually dim the track path ahead of the current position


ksalamy:
<img width="1462" height="544" alt="Image" src="https://github.com/user-attachments/assets/420d2439-a79a-463b-9440-ae4df62b518e" />

## Steve Haddock request

> I can sort of see the points disappearing along the track, but is there supposed to be another visual indicator of where you are in the timeline scrubbing? I guess I expected the path to also dim...

## Implemented behavior (confirmed with product owner)

When the user scrubs the mission progress timeline, the vehicle GPS track splits at the scrub position:

- **Past segment** (already traveled, left of scrub): solid line at reduced opacity (de-emphasized — the vehicle has been here)
- **Future segment** (still to navigate, right of scrub): bright dashed line (emphasized — this is where the vehicle is headed)

This gives a clear visual reading: the dim solid tail shows history, the bright dashed line shows the upcoming path.

*Note: original issue text was less specific about which segment to dim. The dim-past / bright-future treatment was confirmed by the product owner during review.*

PR #659, opened 2026-06-02
refine(#652): invert track emphasis and restrict dimming to timeline hover only

ksalamy:
## Summary

Follow-up refinements to #658 based on user feedback:

- **Inverted visual treatment** — past track is now bright/solid (full opacity); future track is dim (0.5 opacity, weight 2, tight `4, 6` dashes). This matches the "fading trail" convention Steve and Karen both preferred after testing.
- **Timeline-only dimming** — removed `mouseover`/`mouseout` scrub handlers from the invisible map hit circles so hovering the vehicle track on the map no longer triggers the split/dimming effect. The track now only splits when actively hovering the **mission progress timeline bar** at the bottom of the deployment page.

## Test plan

- [ ] Open a vehicle deployment page with a GPS track
- [ ] Hover the mission progress timeline bar — track should split: bright/solid past, dim/dashed future
- [ ] Move mouse off the timeline bar — full solid track restores after ~1 second
- [ ] Hover directly over the vehicle track on the map — confirm no dimming occurs

### DIFF
--- a/apps/lrauv-dash2/components/DeploymentMap.tsx
+++ b/apps/lrauv-dash2/components/DeploymentMap.tsx
@@ -77,6 +77,7 @@ const logger = createLogger('DeploymentMap')
 interface DeploymentMapProps {
   vehicleName?: string | null
   indicatorTime?: number | null
+  dimTime?: number | null
   onScrub?: (time?: number | null) => void
   startTime?: number | null
   endTime?: number | null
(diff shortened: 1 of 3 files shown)


ksalamy:
Updated test plan note: the track split clears after **~250ms** when leaving the timeline bar (that's the `MissionProgressToolbar` mouseout timeout), not ~1s as originally stated in the PR description. The ~1s delay applies to the map track hover dot clearing.

ksalamy:
Re: 250ms vs ~1s (raised again) — this is intentional and already documented in https://github.com/mbari-org/dash5/pull/659#issuecomment-4606261619. The 250ms is the `MissionProgressToolbar` mouseout timeout; no code change needed here.

PR #739, opened 2026-07-01
fix: back button in mission wizard now re-shows Parameters Summary (#723)

### LINKED ISSUE #723: fix: Back button in mission builder skips Parameters summary page
## Summary

In the mission builder, the Forward (Next) and Back (Previous) buttons behave asymmetrically with respect to the interstitial summary pages.

## Steps to Reproduce

1. Open the mission builder and proceed to the Parameters step (step 3)
2. Set one or more parameter overrides
3. Click **Next** — observe that an interstitial Parameters summary page is shown before advancing
4. From the next step (Safety & Comms), click **Back**
5. Observe that the Parameters summary page is skipped entirely — the user lands directly on the Parameters form

## Expected Behavior

Back should mirror Forward — if Next shows a summary interstitial for Parameters, then Back from the following step should also pass through that summary before returning to the Parameters form.

## Additional Feedback

The **Next** and **Back** buttons are positioned at the bottom of the modal. When stepping through multiple screens, the user must scroll or mouse to a new button position each time. Moving these controls to the top of the panel (above the content area) would reduce mouse travel when clicking through steps repeatedly.

## Notes

- Raised by Steve Haddock during Dash5 v5.2.3 review (Jun 2026)
- Root cause is in `useMissionModalSteps.ts` — `handlePrevious` does not account for summary interstitial state when navigating back from a subsequent step
- Priority to be determined

ksalamy:
## Summary

- **Root cause:** `handlePrevious` in `useMissionModalSteps.ts` always left `showSummary=false` when decrementing the step. The forward path correctly sets `showSummary=true` before advancing past Waypoints/Parameters, but there was no equivalent logic on the return path.
- **Fix:** After computing `prevStep`, if landing on a summary-eligible step (Waypoints always; Parameters only when there are overrides), set `showSummary=true` so the summary screen appears before the form — matching the forward flow.
- **No impact on the skip logic:** Steps with no waypoints/parameters/safety still skip correctly on the backward path.

## Test plan

- [x] All 25 existing `MissionModalView` tests still pass
- [x] New regression test: Back from Safety & Comms (step 3) when parameters have overrides → lands on "Summary of overrides" screen, not the Parameters form
- [x] `react-ui` rebuilds cleanly

Closes #723

### DIFF
--- a/apps/lrauv-dash2/components/MissionModal.tsx
+++ b/apps/lrauv-dash2/components/MissionModal.tsx
@@ -85,6 +85,7 @@ const MissionModal: React.FC<MissionModalProps> = ({
     selectedMissionData,
     isRecentRunsLoading: recentRunsLoading,
     isFrequentRunsLoading: frequentRunsLoading,
+    isMissionListLoading,
     missionCategories,
   } = useMissionData({ vehicleName, selectedMission, showAllVehicleMissions })
 
@@ -111,13 +112,11 @@ const MissionModal: React.FC<MissionModalProps> = ({
       return
     }
 
-    // While data is still loading, return early WITHOUT setting the ref so
-    // the effect retries once loading completes. Once loading is done, setting
-    // the ref unconditionally (below) ensures that later re-renders triggered
-    // by user navigation (e.g. Back clearing selectedMission, which causes
-    // missionsWithTemporaryEntry to change) cannot re-run auto-selection and
-    // overwrite the user's chosen tab/category.
-    if (recentRunsLoading || frequentRunsLoading) return
+    // Gate on all three data sources. Returning early (without touching the ref)
+    // lets the effect retry naturally when any loading flag flips to false —
+    // including isMissionListLoading which covers template missions that load
+    // separately from recent/frequent runs.
+    if (recentRunsLoading || frequentRunsLoading || isMissionListLoading) return
 
     if (
       missionPath &&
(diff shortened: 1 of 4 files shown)


ksalamy:
**Additional fix added:** The category dropdown (Recent Runs / Frequent Runs / Templates) was resetting to the wrong tab when clicking Back.

**Root cause:** `hasAutoSelectedRef.current` was only set to `true` when a matching mission was found in the auto-select `useEffect`. If missions data hadn't loaded yet on modal open, the ref stayed `false`. Clicking Back triggered `setSelectedMission(undefined)` → `missionsWithTemporaryEntry` recomputed → `useEffect` fired again → this time it found a mission (e.g. from a Templates category) and overwrote the user's chosen tab with that category.

**Fix:** Set `hasAutoSelectedRef.current = true` unconditionally the first time `missionsWithTemporaryEntry` has data, so any subsequent re-renders caused by Back navigation can never re-trigger auto-select.

ksalamy:
Addressed Copilot's review — updated the auto-select logic so the loading state determines when the ref is set, not whether a match was found:

- **While `recentRunsLoading || frequentRunsLoading` is true:** effect returns early without touching the ref, so it retries naturally once loading completes. ✅ (Copilot's concern)
- **Once loading is done:** `hasAutoSelectedRef.current` is set unconditionally before the match lookup. Any subsequent Back-navigation re-renders (which change `missionsWithTemporaryEntry`) are blocked by the ref. ✅ (original tab-reset bug)

PR #741, opened 2026-07-01
fix: preserve clipboard state between Ctrl+C/V in doc editor (#740)

### LINKED ISSUE #740: bug: Ctrl+C / Ctrl+V second paste inserts wrong content in doc editor
## Reported by
Monique Messié — Jun 30, 2026

## Description
When editing documents, Ctrl+C / Ctrl+V misbehaves: the second paste inserts something other than the content that was just copied.

## Root cause
`DocEditorTipTap.tsx` syncs the `html` prop back into the TipTap editor via a `useEffect` that calls `editor.commands.setContent(html)` whenever `editor.getHTML() !== html`. TipTap's HTML serializer can produce subtly different output from what was stored in React state (attribute ordering, whitespace normalisation), so the comparison returns `true` even immediately after an internal edit — triggering an unnecessary `setContent` call. That call resets ProseMirror's internal clipboard state, so a subsequent Ctrl+V pastes stale or unexpected content instead of what was just copied.

## Fix
Track the last HTML string produced by the editor itself in a ref (`editorHtmlRef`). In the `useEffect`, only call `setContent` when the incoming `html` prop differs from the ref — meaning the change is genuinely external (loading a new document or duplicate). Edits made within the editor skip the `setContent` call entirely, preserving ProseMirror's clipboard state.

ksalamy:
## Summary

- **Root cause:** `DocEditorTipTap` called `editor.commands.setContent(html)` in a `useEffect` whenever `editor.getHTML() !== html`. TipTap's HTML serializer can produce subtly different output from what React stored (whitespace, attribute ordering), so this comparison would return `true` even immediately after an internal user edit — triggering an unnecessary `setContent`. Each `setContent` call resets ProseMirror's internal clipboard state, so a subsequent Ctrl+V pasted stale or wrong content instead of what was just copied.
- **Fix:** Added `editorHtmlRef` to track the last HTML string produced by the editor's own `onUpdate`. The `useEffect` now only calls `setContent` when `html` differs from that ref — meaning the incoming change is genuinely external (e.g. loading a new document or duplicate). Internal edits skip the `setContent` call entirely, keeping ProseMirror's clipboard intact.

## Test plan

- [x] Open a document in edit mode
- [x] Copy text with Ctrl+C, paste with Ctrl+V — first paste works
- [x] Paste again with Ctrl+V — should paste the same copied content, not something different
- [x] Load a new document while editor is open — content should sync correctly (external load still works)
- [x] Duplicate a document — editor should populate with the duplicated content correctly

Closes #740

### DIFF
--- a/apps/lrauv-dash2/components/MissionModal.tsx
+++ b/apps/lrauv-dash2/components/MissionModal.tsx
@@ -121,9 +121,13 @@ const MissionModal: React.FC<MissionModalProps> = ({
       const matchingMission = missionsWithTemporaryEntry.find(
         (m) => m.id === missionPath || m.missionPath === missionPath
       )
-      // Only auto-select once per modal open
+      // Mark as auto-selected now (before checking matchingMission) so that
+      // any subsequent re-render triggered by state changes (e.g. clearing
+      // selectedMission on Back) doesn't re-run the auto-select and overwrite
+      // the user's tab/category choice.
+      hasAutoSelectedRef.current = true
+
       if (matchingMission) {
-        hasAutoSelectedRef.current = true
         setSelectedMission(matchingMission.id)
         // If rerunning from schedule history (has eventData), always use 'Recent Runs'
         if (eventData) {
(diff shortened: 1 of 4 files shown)


ksalamy:
Closing in favour of a clean PR — the original branch accidentally included mission wizard commits from #739. A new branch (fix/doc-editor-clipboard-ctrl-v-clean) rebased onto develop contains only the clipboard fix + regression tests.

PR #759, opened 2026-07-06
feat: add quick-pick confirmation for deployment start date

### LINKED ISSUE #757: fix: clamp deployment start time to now when querying vehicle GPS fixes
## Summary

If a deployment's **Start** timestamp is set to a future date, Dash5 uses that future date as the `from` parameter when querying GPS fixes via the `vehiclePos` API. This returns 0 fixes and causes the vehicle to disappear from the map entirely — even if the vehicle is actively deployed and reporting positions right now.

## Reproduction (Pontus 67, July 6 2026)

- Deployment Start set to **July 31, 2026** (25 days in the future)
- Launch / Vehicle in water: **July 6, 2026** (actual deploy date)
- Dash5 queries GPS fixes `from=1785519960000` (July 31) → 0 results → vehicle not shown on map
- Dash4 uses a different lookback approach → shows pontus correctly

## Root cause

This is a user error (wrong date entry) that Dash5 does not guard against at input time.

## Proposed fix — confirmation picker at save time (no banners)

Replace the free-form Start date field with a small confirmation step that offers sensible quick-pick options before committing:

**Quick-pick options:**
- **Now** — stamps the current time
- **In 1 hour** — for imminent deployments
- **Pick a date** — opens a date/time picker for custom entry

This mirrors the existing "Launch / Vehicle in water" and "Vehicle recovered" button pattern which stamps the current time on click.

If the user selects a future date via the picker, show a single inline note (not a banner) directly below the field:
> *"Start is set to [date] — no GPS fixes will appear until then."*

This makes the consequence of a future date explicit without blocking the save or adding map-level banners.

## Explicitly out of scope

- Silent clamping (`min(startTime, now)`) — masks data errors and could break legitimately pre-planned deployments
- Map-level warning banners — too noisy, not the right place for data entry validation

## Next steps

- [ ] Update `DeploymentDetailsPopUp` in `@mbari/react-ui` to replace the Start date input with the confirmation picker pattern
- [ ] Add inline note when a future date is selected
- [ ] Update `DeploymentDetails.tsx` in the app to wire up the new UX

ksalamy:
## Summary

Replaces the unguarded free-form \`DateField\` for the deployment **Start** date with a simplified quick-pick flow, preventing the user error where a future date is accidentally saved (causing the vehicle to disappear from the map).

## Start date edit flow

When the user clicks the edit (pen) icon and edits the Start row, two options are shown:

| Option | Behaviour |
|---|---|
| **Now** | Saves immediately via \`updateDeployment\` and exits edit mode — giving instant visual feedback that the time was captured |
| **Custom date…** | Opens the existing \`DateField\` with a ← Back link; user reviews the date and clicks **Save Changes** to commit. Clicking ← Back discards the typed value so it cannot be silently saved from the quick-pick view. |

**\"In 1 hour\" was intentionally excluded.** A future-dated preset would recreate the exact problem this feature is solving (vehicles disappearing from the map). The only preset offered is the unambiguous current time.

**\"Now\" saves immediately by design.** Unlike \"Custom date…\", clicking \"Now\" is unambiguous — there is nothing to review. Saving and closing immediately gives clear visual feedback (the display mode updates with the saved timestamp). Waiting for a separate \"Save Changes\" click left the UI in a silent, ambiguous pending state per user feedback.

## Behaviour change: \"Mark start time now\" in display mode

When no start date is set yet, the existing \"Mark start time now\" button now calls \`onSaveChanges\` (i.e. \`updateDeployment\`) instead of \`onSetDeploymentEventToCurrentTime\` (i.e. \`alterDeployment\`). This is a **required fix** — \`alterDeployment\` only accepts \`'launch' | 'recover' | 'end'\` and returns a 400 error for \`'start'\`. The new routing is the correct API call for persisting a start date.

If the saved start date is in the future (display mode, or after entering a custom date), an inline italic note appears:
> *Start is in the future — GPS fixes won't appear until then.*

## What is unchanged

- Launch, Recover, and End date fields — completely unchanged, still use the standard \`DateField\`

## Test plan

- [ ] Click pen icon → Start row shows **Now** and **Custom date…** buttons only
- [ ] Click **Now** → edit panel closes, display mode shows the saved timestamp (no 400 error)
- [ ] Click **Custom date…** → \`DateField\` appears with **← Back** link
- [ ] Click **← Back** → custom date is discarded, returns to quick-pick buttons
- [ ] Enter a past date via Custom → saves without warning
- [ ] Enter a future date via Custom → amber inline note appears
- [ ] Click **Cancel** or **X** then reopen edit mode → starts at quick-pick, not the DateField
- [ ] Deployment with existing future start date → amber note visible in display mode
- [ ] No start date set → \"Mark start time now\" saves via \`updateDeployment\` (no 400 error)
- [ ] Launch / Recover / End rows still use the standard \`DateField\` edit flow

Closes #757

### DIFF
--- a/packages/react-ui/src/PopUps/DeploymentDetailsPopUp.test.tsx
+++ b/packages/react-ui/src/PopUps/DeploymentDetailsPopUp.test.tsx
@@ -107,10 +107,10 @@ test('should display quick-pick buttons for start when edit dates is clicked', a
   fireEvent.click(screen.getByLabelText(/edit dates button/i))
 
   expect(screen.getByLabelText(/set start time to now/i)).toBeInTheDocument()
-  expect(
-    screen.getByLabelText(/set start time to one hour from now/i)
-  ).toBeInTheDocument()
   expect(screen.getByLabelText(/pick a custom start date/i)).toBeInTheDocument()
+  expect(
+    screen.queryByLabelText(/set start time to one hour from now/i)
+  ).not.toBeInTheDocument()
   expect(screen.queryByLabelText(/edit dates button/i)).not.toBeInTheDocument()
 })
 
--- a/packages/react-ui/src/PopUps/DeploymentDetailsPopUp.test.tsx
+++ b/packages/react-ui/src/PopUps/DeploymentDetailsPopUp.test.tsx
@@ -148,6 +148,35 @@ test('should show inline warning when existing start date is in the future', asy
   expect(screen.getByText(/Start is in the future/i)).toBeInTheDocument()
 })
 
+test('should show inline warning in quick-pick view when existing start date is in the future', async () => {
+  const futureDate = new Date(
+    Date.now() + 7 * 24 * 60 * 60 * 1000
+  ).toISOString()
+  render(<DeploymentDetailsPopUp {...props} startDate={futureDate} />)
+
+  fireEvent.click(screen.getByLabelText(/edit dates button/i))
+
+  // quick-pick buttons should be visible and warning should appear below them
+  expect(screen.getByLabelText(/set start time to now/i)).toBeInTheDocument()
+  expect(screen.getAllByText(/Start is in the future/i).length).toBeGreaterThan(
+    0
+  )
+})
+
+test('should show inline warning in custom date view when start date is in the future', async () => {
+  const futureDate = new Date(
+    Date.now() + 7 * 24 * 60 * 60 * 1000
+  ).toISOString()
+  render(<DeploymentDetailsPopUp {...props} startDate={futureDate} />)
+
+  fireEvent.click(screen.getByLabelText(/edit dates button/i))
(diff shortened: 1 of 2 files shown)


ksalamy:
**Responding to Copilot's latest review:**

**"In 1 hour" option**: Intentionally removed. After discussion with the team, the "In 1 hour" preset was considered a shortcut that could still result in an incorrect future start date (the same problem we're trying to prevent). The simplified flow is deliberately **Now** or **Custom date…** only.

**"Now" saves immediately**: Also intentional. The previous request was specifically for visible feedback when "Now" is clicked — the user found that updating local state silently (and only persisting on "Save Changes") was confusing. Clicking "Now" is unambiguous, so we persist and close immediately, mirroring the "Mark start time now" behavior in display mode. "Custom date…" still follows the review-then-save pattern. The tests accurately reflect this intended design.

PR #764, opened 2026-07-08
fix: preserve depth chart zoom state across hover-driven re-renders

### LINKED ISSUE #763: bug: depth chart zoom resets immediately when mouse moves after zooming in
## Description

Reported by Akshay Hinduja and confirmed by Emme Wiederhold.

In the **Depth Data** tab (next to Vehicle State), zooming into a region of the depth graph immediately reverts to the original scale as soon as the mouse is moved (~2 seconds). Zooming **out** appears to stick correctly.

This behavior is not observed in the Science and Vehicle Data tab charts — only the Depth Data tab.

## Steps to reproduce

1. Open any vehicle deployment
2. Click the **Depth Data** tab (next to Vehicle State)
3. Attempt to zoom into a region of the depth graph (click+drag to crop/zoom)
4. Move the mouse → chart immediately snaps back to original scale

## Expected behavior

Zoom should persist until the user explicitly resets it (e.g. double-click or reset button).

## Likely cause

The Depth Data chart is almost certainly on a short refetch/polling interval. When new data arrives, the chart re-renders and resets the zoom domain back to the default. The Science/Vehicle Data charts may handle this differently (e.g. preserving user-set zoom domain across re-renders) or have a longer polling interval.

## Reported by

- Akshay Hinduja
- Emme Wiederhold

ksalamy:
## Summary

Fixes the depth chart zoom immediately resetting when the user moves the mouse after zooming in (reported in the Depth Data tab only, not Science/Vehicle Data charts).

**Root cause:** The Depth Data chart is the only `LineChart` consumer that wires `onHover` to a parent `setState` (`setIndicatorTime`). Every mouse move triggered a re-render, which passed a fresh `data` array reference to Plotly. Without `uirevision`, Plotly treats each new render as a full reset and discards the user's zoom state.

## Changes

- **`LineChart.tsx`** — Added an opt-in `uirevision` prop. When provided and unchanged across renders, Plotly preserves zoom/pan state; when it changes, Plotly resets the axes. When omitted (default), Plotly's standard reset-on-render behavior is preserved for all other consumers.
- **`DepthSection.tsx`** — Passes `uirevision` keyed to `vehicleName`, deployment `from`, time window, and logset ID so zoom persists across hover re-renders but resets correctly when switching vehicles, deployments, time windows, or logsets. Also memoizes chart data points, clamping to the shorter of `values`/`times` to guarantee in-bounds, defined timestamps.

Science/Vehicle Data charts are unaffected — they don't wire `onHover` to parent state and don't pass `uirevision`.

## Test plan

- [ ] Navigate to the Depth Data tab for any active vehicle deployment
- [ ] Zoom into a region of the depth chart
- [ ] Move the mouse — zoom should now stay put
- [ ] Verify the timeline scrubber/indicator still updates as the mouse moves over the chart (hover wiring is unchanged)
- [ ] Switch time window (e.g. Latest Dive -> 3 Days) — zoom should reset
- [ ] Navigate to a different vehicle/deployment — zoom should reset
- [ ] Confirm Science/Vehicle Data charts still work normally

Closes #763

### DIFF
--- a/apps/lrauv-dash2/components/DepthSection.tsx
+++ b/apps/lrauv-dash2/components/DepthSection.tsx
@@ -160,6 +160,16 @@ const DepthSection: React.FC<{
     ? depthQuery.data ?? undefined
     : latestChartData?.find((d) => d.name === 'depth')
 
+  const chartPoints = useMemo(
+    () =>
+      depthData?.values?.map((v: number, i: number) => ({
+        value: v,
+        timestamp: depthData.times?.[i],
+      })),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [depthData?.values, depthData?.times]
+  )
+
   // While logsets are still loading / auto-selecting, latestQuery is disabled
   // (isLoading = false). Treat that wait as loading so "No depth data" doesn't
   // appear prematurely before any data has been fetched.
@@ -222,10 +232,7 @@ const DepthSection: React.FC<{
         {chartAvailable && (
           <LineChart
             name={depthData.name}
-            data={depthData.values?.map((v: number, i: number) => ({
-              value: v,
-              timestamp: depthData.times?.[i],
-            }))}
+            data={chartPoints}
             yAxisLabel={`${humanize(depthData.name)} (${depthData.units})`}
             onHover={onHover}
             inverted
--- a/apps/lrauv-dash2/components/DepthSection.tsx
+++ b/apps/lrauv-dash2/components/DepthSection.tsx
@@ -162,14 +162,16 @@ const DepthSection: React.FC<{
 
   const depthValues = depthData?.values
   const depthTimes = depthData?.times
-  const chartPoints = useMemo(
-    () =>
-      depthValues?.map((v: number, i: number) => ({
-        value: v,
-        timestamp: depthTimes?.[i],
-      })),
-    [depthValues, depthTimes]
-  )
+  const chartPoints = useMemo(() => {
+    if (!depthValues || !depthTimes) return undefined
+    // Clamp to the shorter array so indices are always in-bounds and
+    // timestamps are always defined (both arrays are number[] per the API).
+    const len = Math.min(depthValues.length, depthTimes.length)
+    return Array.from({ length: len }, (_, i) => ({
+      value: depthValues[i],
+      timestamp: depthTimes[i],
+    }))
+  }, [depthValues, depthTimes])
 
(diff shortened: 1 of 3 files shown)


ksalamy:
Thanks for the review! Both comments are addressed:

**Comment 1 (uirevision keyed only to name):** Already resolved in the follow-up commit — `uirevision` was made a prop on `LineChart` (defaulting to `name` for other consumers). `DepthSection` now passes `uirevision={`depth-${timeWindow}-${selectedLogsetId}`}`, so zoom is stable across hover re-renders but resets correctly when the time window or logset dropdown changes.

**Comment 2 (eslint-disable in useMemo):** Fixed — extracted `depthValues` and `depthTimes` as local variables before the `useMemo` so the dependency array can reference them directly without suppressing the lint rule.

ksalamy:
All three comments addressed:

1. **uirevision opt-in** — Changed `uirevision ?? name` to `...(uirevision !== undefined && { uirevision })`. The layout key is now only set when the caller explicitly passes the prop, so existing LineChart consumers keep Plotly's default reset behavior unchanged.

2. **JSDoc updated** — Now clearly states opt-in semantics and that omitting the prop restores Plotly's default behavior.

3. **Unit tests added** — Mocked `react-plotly.js` to capture the `layout` prop and added two tests: one asserting `uirevision` is absent when the prop is omitted, one asserting it equals the provided value. All 3 tests pass.

ksalamy:
Fixed — switched to `'uirevision' in layout` so the mock correctly enforces key absence rather than just checking for `undefined`. The test now fails if `LineChart` ever starts setting `uirevision: undefined` explicitly. All 3 tests still pass.

ksalamy:
Fixed — `uirevision` now includes `vehicleName` and `from` (the deployment start boundary): ```\n`${vehicleName}-${from}-${depthData.name}-${timeWindow}-${selectedLogsetId}`\n```\nNavigating to a different vehicle or deployment now produces a different key and correctly resets zoom.

ksalamy:
Both comments addressed:

1. **PR description mismatch** — Updated the description to accurately reflect that `uirevision` is an opt-in prop (not a hardcoded `name` default), and that `DepthSection` passes a compound key scoped to vehicle, deployment, time window, and logset.

2. **`depthTimes?.[i]` undefined risk** — Replaced with an explicit `Math.min` length guard and `Array.from` so both arrays are always accessed in-bounds and timestamps are always `number`, eliminating the epoch fallback entirely.

PR #781, opened 2026-07-17
replace native confirms with custom dialog

### LINKED ISSUE #683: refactor: replace native confirm() dialogs with a proper confirmation modal
## Background

Five places in the app use the native browser \`confirm()\` / \`window.confirm()\` to ask the user before a destructive action. This is a React anti-pattern:

- **Blocks the JavaScript main thread** — nothing can update or render while the dialog is open.
- **Cannot be styled** — looks completely out of place next to the rest of the app's UI.
- **Not accessible** — no focus management, no keyboard trap, does not integrate with ARIA.
- **Can be suppressed** — some browsers and environments (embedded webviews, test runners) block or auto-dismiss \`confirm()\`.

## Current usages

| File | Line | Message |
|---|---|---|
| \`components/EmailNotificationsModal.tsx\` | 542 | \`Delete all notification settings for <email>?\` |
| \`components/EmailNotificationsModal.tsx\` | 605 | \`Remove <email> from your notification addresses?\` |
| \`components/MarkerContext.tsx\` | 328 | \`Are you sure you want to remove all markers? This cannot be undone.\` |
| \`components/MarkerSection.tsx\` | 76 | \`Remove this marker from map layers?\` |
| \`components/ScheduleSection.tsx\` | 1268 | \`Cancel this <type> directive (event ID <id>)? This will remove it from the shore-side queue.\` |

## Proposed Fix

Add a small, reusable \`ConfirmDialog\` component (or a \`useConfirm\` hook) to \`apps/lrauv-dash2/components/\` that renders a lightweight inline confirmation modal. Replace all five \`confirm()\` call sites with it.

The confirmation UI only needs:
- A short message/question (passed as a prop)
- A **Confirm** button (destructive/danger style) and a **Cancel** button
- Focus trap while open
- Closes on Escape

The existing \`Modal\` primitive from \`@mbari/react-ui\` can be used as the base.

## Acceptance criteria
- [ ] No \`confirm()\` or \`window.confirm()\` calls remain in the codebase
- [ ] All five destructive actions above show a styled, in-app confirmation dialog before proceeding
- [ ] Cancel aborts the action; Confirm proceeds
- [ ] Dialog closes on Escape
- [ ] No new lint or type errors

## Notes
- No functional change — this is purely a UX and code quality improvement.
- Good first issue: the scope is fixed and each replacement follows the same pattern.

ewiederhold:
this should fulfill #683; I created a new provider and custom `useConfirm` hook, where the provider leverages the existing `Modal` component as suggested in the issue

I replaced all five of the native `window.confirm` calls that were listed in the ticket, _**however**_ please note the comments below about unused code because I believe this code should be modified or removed before we merge this PR so as to not add more unused code to the repo

lastly, I chose not to implement the "close on escape" request because the modal already contains two means of cancelling (an "X" in the upper right and a "Cancel" button), and I couldn't find any other dialog modals in the app that allow using the escape key to cancel so this would have been a diversion from typical functionality in our other existing components

### unused code
- it seems the `clearAllMarkers` callback defined in `MarkerContext.tsx` is never used anywhere in the codebase, should we remove it? I couldn't find any way to test my changes for that component on the UI
- looks like the `MarkerSection.tsx` component isn't used anywhere in the codebase either, also couldn't find any way to test my changes to that component – should we delete it?


ksalamy (changes requested):
PR Review: refactor/confirm-modals

**Overall: solid first PR.** Clean, well-structured approach. 

**What works well**
1. The overall approach is the right one. Replacing window.confirm() with a Promise-based React context dialog is the correct pattern for a modern React app — window.confirm() is synchronous, blocks the thread, and can't be styled. This is a meaningful improvement.
2. ConfirmationProvider is placed at the right level in _app.tsx — wrapping everything so all consumers can reach it.
3. The useConfirm hook throws a clear error if used outside the provider, which makes misuse obvious at runtime rather than silently returning undefined.
4. The confirmButtonAppearance addition to Modal / Footer is a clean, reusable extension — that prop will be useful beyond this PR.
5. The migration across EmailNotificationsModal, MarkerContext, MarkerSection, and ScheduleSection is consistent and the call sites read cleanly.

**Issues**
1. **Critical Bug** — Cancel still executes the directive cancellation In ScheduleSection.tsx, the queryClient.invalidateQueries(), toast.success(), and cancellation note logic sit outside the if (isConfirmed) block, after the closing brace. They run unconditionally regardless of which button the user clicked. Confirmed with live testing: hitting "Cancel" on the dialog still cancels the directive. Everything after the if (isConfirmed) { ... } block needs to move inside it.

2. **React anti-pattern** — resolver in useState Storing the Promise resolver in useState with setResolver({ resolve }) works only because it's wrapped in an object (if you ever pass resolve directly, React calls it as an updater and invokes it immediately — silent breakage). The conventional and safe approach is useRef:

```tsx
const resolverRef = useRef<((value: boolean) => void) | null>(null)
// in confirm():
return new Promise<boolean>((resolve) => { resolverRef.current = resolve })
// in handleConfirm/handleCancel:
resolverRef.current?.(true)
```

This avoids both the anti-pattern and the fragility of the object wrapper.


3. **UX**

- Confusing button labels on the cancel-directive dialog The confirm dialog title reads: "Cancel this mission directive (event ID …)? This will remove it from the shore-side queue." The buttons shown are Confirm and Cancel. Since the action being confirmed is itself a cancellation, "Cancel" means "don't cancel" — which is genuinely confusing in practice. Consider custom labels using the already-supported confirmLabel and cancelLabel options: e.g. confirmLabel: 'Yes, cancel directive' and cancelLabel: 'Keep' or 'Go back'. These fields exist on ConfirmOptions but are never passed at any call site in this PR.
- Main modal closes immediately after "Delete all" In EmailNotificationsModal, after the user confirms "Delete all notification settings", onClose?.() is called directly, dismissing the entire modal with no visual feedback. A toast or inline success state should appear before or instead of closing, so the user knows the deletion succeeded. (The "Remove email address" flow handles this correctly — it closes only the sub-dialog and returns to the main modal with the primary email selected.)

4. **No tests for ConfirmContext**
No tests for ConfirmContext The new provider and hook have no unit tests. At minimum, tests should verify: (a) confirm() resolves true when the confirm button is clicked, (b) resolves false when cancel/close is clicked, and (c) useConfirm throws when used outside the provider.


**NOTE:** Pre-existing gap (not introduced by this PR) The marker-related confirm dialogs in MarkerContext and MarkerSection cannot currently be reached in the UI because MarkerSection is orphaned from the navigation (I have added this bug separately in Issue [#782](https://github.com/mbari-org/dash5/issues/782)). The confirm migration there is correct — it just can't be tested until that gap is closed.  _I have assigned myself to tackle this part of the code._

**One suggestion for future PRs - *this was suggested to me as well when I began PRs* **
Before requesting human review, consider running a Copilot review pass on the PR. It's particularly good at catching React anti-patterns (like the useState / useRef issue above) and logic bugs like the unconditional post-if block. Running it first means the human review can focus on design decisions, UX intent, and broader behavior — which is the more valuable conversation to have together.

ewiederhold:
@ksalamy
1. good catch, I've pushed a commit to fix that
2. I acknowledged that while I was writing these changes, but left it the way it is because the resolver is only ever set within the provider where it currently is (and uses the correct bracket notation) – I don't expect any other file to call this provider's resolve and so the object notation should be fine in this one place and works as expected
3. making the confirm modal buttons have custom text would require some rework, and I'd rather rephrase the question being asked of the user than change the button convention for only one of the modals (if users are accustomed to "Cancel" and "Confirm," changing those in one of the modals might make it even more confusing and lead to accidental confirmation of things)
   - for the second bullet, I described this exact issue in #783 already; this is not something introduced by this PR, it's an existing bug in prod
4. can you give an example of writing tests for something like this? I couldn't find any documentation in the repo about how we write tests, when we should write tests, best practices, etc.

and noted on triggering a copilot review first next time, but I also don't see instructions for that anywhere, can you point me to some? any other PR best practices you'd like me to keep in mind?

ksalamy:
Thanks,
1. **Critical bug (ScheduleSection):** Looks good on re-check — Cancel no longer cancels the directive.
2. **UX:** I’ll take **rephrasing the question** over custom buttons as long as it clarifies things to the user. `#783` covers the delete-all modal close.
3. **Resolver in `useState`:** The object wrapper works, but `useRef` is safer long-term — small change and worth it for future work.
4. **Tests:** Dash5 doesn’t have a dedicated testing guide yet. Closest examples:
   - `apps/lrauv-dash2/components/MarkerContext.test.tsx`
   - `apps/lrauv-dash2/components/MarkersLayerSection.test.tsx`
  
   For `ConfirmContext`, the useful minimum is: resolve `true` on confirm, resolve `false` on cancel/close, and throw when `useConfirm` is used outside the provider.

**Copilot:** Trigger by adding Copilot to the reviewers list (top right of the PR) before requesting human review.

ksalamy:
@ewiederhold — circling back on #781. The critical Cancel fix looks good. Still waiting on the useRef + tests from my Jul 18 note when you have a chance. Thanks!

ewiederhold:
@ksalamy requested updates are complete, though these changes cause the `ScheduleSection` tests to fail – I would update them, but there are two suites, one correctly defined in `/jest` and one incorrectly in `/components` – combining these tests seems outside the scope of this PR, while updating them to be fixed could be in the scope of this PR since this PR's changes break those tests... let me know how to proceed

ewiederhold:
@ksalamy I understand, but the `ScheduleSection.test.tsx` in the `components` directory **_doesn't_** seem to be an exact duplicate of the one in `jest`, which is why I was wondering how to proceed – I don't think the `components` one should just be outright deleted without considering some of its tests to be pulled over to `jest`

do you want me to make sure that both of them run successfully for this PR? even though it's the work of #815 to combine and delete some of those tests?

ksalamy:
Hi @ewiederhold I see what you mean. They are not exact duplicates...... I created an issue to fix test issues (#815) a few days ago (consolidate older /components cases into the /jest and then deleting the dup). You don't need to do that on this PR. 

Both test suites pass locally on my machine. The Github CI is queued on your last commit. I will watch it and approve this PR once it is green. You should not need to confirm CI or re-request.

PR #787, opened 2026-07-18
fix(#782): restore Markers layer actions and persistence

### LINKED ISSUE #782: Markers layer: restore clear/remove UI and fix layer persistence
## Summary

The `MarkerSection` component's remove and clear-all buttons are unreachable because `MarkerSection` is no longer rendered anywhere in the app after the `MapLayersListModal` refactor.

**Note:** Individual markers can still be removed by clicking them directly on the map and using the trash icon. What is missing is the list-based marker management UI (`MarkerSection`) and specifically layer list actions (remove from layer / remove all from layer), which have no equivalent in the current Layers modal UI.

## Fix approach

Wire the missing actions into `MarkersLayerSection` (the live Layers modal section) and remove the orphaned `MarkerSection.tsx`.

## Persistence bug (same issue)

Markers saved to the Map Layer are supposed to persist across refresh via `localStorage` (`lrauv-map-markers`). They often do not.

### Cause

In `MarkerContext.tsx`, mount-time effects race:

1. Initial `markers` state is `[]`
2. A `useEffect` immediately writes `markers` to `localStorage` — wiping stored layer markers to `[]`
3. An async `getMarkers()` fetch can then reload that empty array and overwrite in-memory state

There is also conflicting write logic: some paths persist only `savedToLayer` markers, while another effect persists the entire `markers` array on every change.

### Expected

- Markers with **Save to Map Layers** survive page refresh and still appear under Layers → Markers
- Unsaved (session-only) markers may clear on refresh; layer-saved markers must not

## Center-on for layer markers (same issue)

Stations and polygons in the Layers modal support centering the map on an item. Layer-saved markers should get the same control (fly-to using marker lat/lng).

## Marker color parity in Layers list (same issue)

If a user changes a marker’s color on the map and then saves it to the Map Layer, the Layers modal should show that same color (via `iconColor`) instead of always rendering a hardcoded red marker icon.

## UI clarifications

- Layers list unlink = remove from layer (marker stays on map)
- Map popup trash = permanent delete (map + layer)
- Checkbox = show/hide on map (stays in layer)
- Center control = fly map to that marker (parity with stations/polygons)
- Layers list icon color = match the marker’s map color (`iconColor`)
- Delete orphaned `MarkerSection.tsx` once behavior lives in `MarkersLayerSection`

### LINKED ISSUE #784: Map Layers should display correct Marker colors
The map Layers UI could be easier to quickly navigate by making the marker icon's to the left of the markers match the color that the corresponding marker is on the map

The markers that are saved to the map layers should have the correctly colored icon next to them in the map layer selection menu instead of all using the default red color

note below that the marker is green on the map, but red in the layers modal
Image

ksalamy:
Review Ready.

Review order: Copilot first, then human reviewers (requested when clean).

## Summary
- Restore Markers list actions in the Layers modal: remove from layer, remove all from layer, center on marker, and per-marker icon color
- Fix layer-marker persistence so markers saved to the Map Layer survive page refresh
- Remove orphaned `MarkerSection.tsx` (behavior now lives in `MarkersLayerSection`)

Closes #782
Closes #784

## Notes
- Layers unlink removes from the layer only; the marker stays on the map
- Map popup trash remains the permanent delete path
- `window.confirm` is used for layer-remove confirms (same pattern as existing marker code)
- Marker color parity in the Layers list (`iconColor`) addresses #784

## Test plan
- [ ] Save markers to Map Layers → they appear under Layers → Markers with matching colors
- [ ] Hard refresh → layer-saved markers persist on the map and in Layers
- [ ] Unsaved session marker does not persist after refresh
- [ ] Unlink one marker → stays on map, leaves Layers
- [ ] Remove all from layer → markers stay on map, Layers empty
- [ ] Center icon flies the map to that marker
- [ ] Map popup trash permanently deletes the marker
- [ ] `yarn workspace @mbari/lrauv-dash2 test:ci -- components/MarkersLayerSection.test.tsx components/MarkerContext.test.tsx`


ksalamy:
Note: also closes #784 (Layers marker icon color parity), which was filed separately and is already implemented here.

apps/lrauv-dash2/components/MapLayersTreeItem.tsx
```diff
@@ -212,6 +222,43 @@ export const TreeItem: React.FC<TreeItemProps> = ({
...
+                  display: 'inline-flex',
+                  alignItems: 'center',
+                  justifyContent: 'center',
+                  cursor: 'pointer',
+                }}
+              >
+                <FontAwesomeIcon
+                  icon={removeIcon}
+                  style={{ color: '#6b7280', fontSize: '13px' }}
+                />
```
ewiederhold, line 258: I know this isn't necessarily new with this PR, but I'm surprised to see hard-coded hex values for colors instead of referencing a global theme or something more consistent?

apps/lrauv-dash2/components/MarkerContext.tsx
ewiederhold, line 1: at the top of this file, there's an import for api helpers from `@mbari/api-client` that's commented out "temporarily" – why aren't we using those utility functions for CRUD of markers instead of storing everything locally?

additionally, from a UX perspective, it feels like markers maybe shouldn't disappear when the user reloads the page (even the markers not saved to layer)? as in, the markers saved to layer should probably exist outside of local storage and use the API for CRUD operations, and then local storage can be used for other markers a user is experimenting with but that aren't saved to the backend via the API – I'd say this is yet another thing to take the team's temperature on before making big code changes that we might have to undo

apps/lrauv-dash2/components/MarkerContext.tsx
```diff
@@ -59,32 +59,52 @@ export interface MarkerContextType {
...
-  return savedMarkers ? JSON.parse(savedMarkers) : []
+const readStoredLayerMarkers = (): MarkerData[] => {
+  if (typeof window === 'undefined') return []
+  try {
+    const raw = localStorage.getItem(STORAGE_KEY)
+    if (!raw) return []
+    const parsed = JSON.parse(raw) as MarkerData[]
+    if (!Array.isArray(parsed)) return []
+    // Only restore markers explicitly saved to the layer (ignore legacy session entries)
+    return parsed.filter((marker) => marker.savedToLayer === true)
```
ewiederhold, line 85: this is the part I'm debating above in my file-level-comment about whether we should actually be clearing "legacy session entries" (unless the user does a hard refresh/clear local cache)

apps/lrauv-dash2/components/MarkerContext.tsx
```diff
@@ -441,7 +409,7 @@ export const MarkerProvider: React.FC<{ children: React.ReactNode }> = ({
       const marker = markers.find((m) => m.id === numericId)
       const markerLabel = marker?.label || 'Unnamed'
 
-      // Remove the marker from the array
+      // Remove the marker from the array (persist effect updates localStorage)
```
ewiederhold: I don't understand the change to this comment, what does "persist effect updates localStorage" mean?

apps/lrauv-dash2/components/MarkerContext.tsx
```diff
@@ -454,21 +422,6 @@ export const MarkerProvider: React.FC<{ children: React.ReactNode }> = ({
...
-        )
-        try {
-          localStorage.setItem(
-            'lrauv-map-markers',
-            JSON.stringify(updatedMarkers)
-          )
-        } catch (error) {
-          logger.error('Error removing marker from storage:', error)
-        }
-      }
```
ewiederhold, line 472: this logic seems to indicate that we maybe used to use local storage for what I described (making markers to persist locally but that aren't saved to the map layers via the API, but maybe eventually saved to layers at which point they'd be removed from local storage and saved to the backend via the API); did we do it that way in the past? why did we switch?

apps/lrauv-dash2/components/MarkerContext.test.tsx
```diff
@@ -155,4 +171,140 @@ describe('MarkerContext', () => {
...
+      <MarkerProvider>
+        <TestComponent />
+      </MarkerProvider>
+    )
+
+    await userEvent.click(screen.getByTestId('add-marker'))
+    await userEvent.click(screen.getByTestId('clear-all'))
+
+    expect(screen.getByTestId('marker-count').textContent).toBe('1')
+  })
```
ewiederhold: why is this test not combined with the one defined two above it? they're set up in almost an identical way and testing very similar things; is there a reason to have two separate test cases that are near duplicates instead of combining them and having multiple `expect` statements?

apps/lrauv-dash2/components/MarkerContext.tsx
```diff
@@ -313,20 +265,36 @@ export const MarkerProvider: React.FC<{ children: React.ReactNode }> = ({
...
-      return updatedMarkers
+    toast.success('All markers removed from layer', {
+      duration: 3000,
+      className: 'blue-toast',
     })
   }, [])
 
-  // Add function to clear all markers
+  // Permanently delete every marker from the map
   const clearAllMarkers = useCallback(() => {
```
ewiederhold: I'm pretty sure this function still isn't called from anywhere... the function is achieved via `removeAllMarkersFromLayer` instead, so I think we should delete this

apps/lrauv-dash2/components/MarkersLayerSection.tsx
```diff
@@ -1,10 +1,15 @@
...
 
 interface MarkerItem {
   id: number | string
   label?: string
+  lat?: number
+  lng?: number
+  iconColor?: string
   visible?: boolean
   savedToLayer?: boolean
 }
```
ewiederhold: this interface definition is extremely similar to `MarkerData` defined in `MarkerContext.tsx` and they're both interacting with the same markers – why don't we have one single defined interface for markers instead of multiple very similar ones?

apps/lrauv-dash2/components/MarkersLayerSection.tsx
```diff
@@ -41,23 +52,66 @@ export const MarkersLayerSection: React.FC<MarkersLayerSectionProps> = ({
       onToggleCheck={
         layerMarkers.length > 0 ? handleToggleSelectAllMarkers : undefined
       }
+      checkTooltip={
+        layerMarkers.length > 0
+          ? 'Show/hide all markers on map (stay in layer)'
```
ewiederhold: you defined a `VISIBILITY_TOOLTIP` constant within this file for a very similar string, why not for this one?

apps/lrauv-dash2/components/MarkersLayerSection.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+    )
+
+    const checkboxes = screen.getAllByRole('checkbox')
+    // index 0 is the Markers parent checkbox; index 1 is Waypoint A
+    await userEvent.click(checkboxes[1])
+
+    expect(toggleMarkerVisibility).toHaveBeenCalledWith('1')
+    expect(checkboxes[1]).toHaveAttribute(
+      'title',
+      'Show/hide on map (stays in layer)'
```
ewiederhold: maybe import the constant here in case that string changes back in `MarkersLayerSection.tsx` then you don't have to update this hardcoded version – I can imagine an argument for wanting this test to fail if that string constant is changed in the main file, but it doesn't seem necessary to me

apps/lrauv-dash2/components/MarkersLayerSection.tsx
```diff
@@ -41,23 +52,66 @@ export const MarkersLayerSection: React.FC<MarkersLayerSectionProps> = ({
...
-          isChecked={marker.visible !== false}
-          onToggleCheck={() => toggleMarkerVisibility(String(marker.id))}
-          icon={faMapMarkerAlt}
-          iconColor="red"
-        />
-      ))}
+      {filteredMarkers.map((marker) => {
+        const markerLabel = marker.label || `Marker ${marker.id}`
+        const hasValidCoords =
+          Number.isFinite(marker.lat) && Number.isFinite(marker.lng)
```
ewiederhold: instead of checking that these are finite, I would check that they're within our expectations for latitude (+/- 90º) and longitude (+/- 180º) – especially longitude, when outside these bounds, causes the map to behave weirdly, where something like vehicle tracks are only displayed within those bounds of lat and lon, but if a marker is outside those bounds it only displays in the repetition of the map in which that value exists – if a marker is outside those bounds, we should calculate its in-bounds coordinates accordingly before we render it

apps/lrauv-dash2/components/MarkersLayerSection.tsx
```diff
@@ -41,23 +52,66 @@ export const MarkersLayerSection: React.FC<MarkersLayerSectionProps> = ({
...
+            Remove all from layer
+          </button>
+        </div>
+      )}
       {layerMarkers.length === 0 && (
         <div className="py-2 pl-10 text-sm italic text-gray-500">
-          No markers saved to layers
+          No markers saved to layer
         </div>
       )}
```
ewiederhold: why not condense this into `layerMarkers.length > 0 ? (... Remove all from layer ...) : (... No markers saved to layer ...)`?

ewiederhold (changes requested):
more detailed comments below, higher level comment specifically about the "Center map on" function – when the Vehicle State of Depth Data drawers below the map are expanded, the marker center function appears not to work because the center of the map is hidden behind that open drawer on some screen sizes... maybe the marker center functionality should take into account the actual visible map on the screen and use that to calculate the visible center of the map?

apps/lrauv-dash2/components/MapLayersTreeItem.tsx
ksalamy: Yes, the hard-coded hexes came in with the Layers panel…. Currently, there isn’t a shared color theme for the maps (react-ui has Tailwind primary/secondary), but the map features use inline styles for Leaflet, FontAwesome and other picked colors. This could be a separate issue down the road if you feel strongly about it.

apps/lrauv-dash2/components/MarkerContext.tsx
ksalamy: Yes, that is intentional for now. Currently, there is no working backend marker API (client-stubs only). Dash5 keeps layer-saved markers in localStorage so they survive a refresh. Unsaved markers stay in memory only and don’t survive. We will need to revisit this before any redesign. Until then, I have updated the comment with more detail to make things clearer.

apps/lrauv-dash2/components/MarkerContext.tsx
ksalamy: Yes, the filter (marker.savedToLayer === true) is intentional. On load, the browser only restores layer-saved markers. On the next save/persist localStorage is rewritten with those markers and drops any old session-only entries. Users need to _actively_ save the markers they want to see should the browser be refreshed.

apps/lrauv-dash2/components/MarkerContext.tsx
ksalamy: It is/was my way of saying that setMarkers triggers the useEffect that persists layer markers to localStorage.  I have reworded the comment to be clearer.

apps/lrauv-dash2/components/MarkerContext.tsx
ksalamy: Yes, the original Dash5 design _did_ have some hooks built toward saving markers to the layer through an API, but that API path was never finished. For this PR (and for now), localStorage is what keeps the layer-saved markers around across refresh.

apps/lrauv-dash2/components/MarkerContext.test.tsx
ksalamy: These two tests are similar but test the confirm dialog in opposite ways. I felt it was easier to see which path failed if one broke. I can merge them if you prefer a single test. Just let me know.

apps/lrauv-dash2/components/MarkersLayerSection.tsx
ksalamy: I agree, MarkerData and MarkerItem _are_ basically the same....  I have consolidated them both into MarkerData and dropped MarkerItem.

apps/lrauv-dash2/components/MarkersLayerSection.tsx
ksalamy: Done. I have added a constant for the second similar string.

apps/lrauv-dash2/components/MarkersLayerSection.test.tsx
ksalamy: Ok. The test now imports VISIBILITY_TOOLTIP from the source file.

apps/lrauv-dash2/components/MarkerContext.tsx
ksalamy: Agreed. The clearAllMarkers is unused in the app UI (only in tests). It is also different from removeAllMarkersFromLayer (deletes everything vs unlinks from layer). I have removed the function, the context export, and those two tests.

apps/lrauv-dash2/components/MarkersLayerSection.tsx
ksalamy: Agree 100%.  The code now normalizes/clamps coordinates before a center (±90 / wrap ±180).

apps/lrauv-dash2/components/MarkersLayerSection.tsx
ksalamy: These are now condensed into a single ternary.

ksalamy:
Interesting catch. Yes, “Center map on” uses the shared fly-to path (MapFlyTo -> Leaflet setView) to center the map on the full container. As a result, it does not “see” or even know about the Vehicle State/Depth Drawer section. This affects not just the markers but the stations and polygons. This is valid bug and should have its own issue. Follow-up issue - #816.

apps/lrauv-dash2/components/MarkerContext.tsx
ksalamy: New Issue #829 filed to revisit the addition of global markers similar to those in Dash4.

ewiederhold (approved):

Issue #788, opened 2026-07-18
follow-up: gate overview map marker editing behind login


ksalamy:
## Summary

Follow-up to [#786](https://github.com/mbari-org/dash5/issues/786) and implementing PR [#789](https://github.com/mbari-org/dash5/pull/789): gate **overview map marker write actions** behind login, matching the “views open / writes gated at entry” policy.

This was identified during the #786 review as a **residual write path** not covered by that change. #786 focused on documents, schedule, deployment toolbar, and related modal gates. Overview map markers were explicitly left out of scope there.

## Findings

While auditing write entry points for #786:

- On the Overview map (`apps/lrauv-dash2/pages/index.tsx`), markers expose edit / delete / color / position controls (via `DraggableMarker` / marker context) **without an `authenticated` check**.
- Logged-out users can still enter marker-mutating UI (rename, delete, recolor, drag). Even if persistence is mostly client-side today, this breaks the product rule that write affordances should not be available until the user is signed in.
- This is the same class of trap #786 fixed for documents: present a write control → user invests effort → later discover they cannot (or should not) complete the action without login.

## Proposed work

- [ ] Hide or disable marker edit / delete / color / drag (and any “save to layer” style writes) when `authenticated` is false
- [ ] Keep marker **viewing** available while logged out (consistent with #786 view policy)
- [ ] Prefer hide/disable at entry over allowing edits that fail or only persist locally in a confusing way
- [ ] Optional: short “Sign in to edit map markers” affordance if the empty/missing controls feel unclear

## Out of scope

- Re-opening the broader #786 document/schedule/deployment gates (handled separately)
- App-wide must-login redirect

## References

- Parent: [#786 — gate write actions behind login (views stay open)](https://github.com/mbari-org/dash5/issues/786)
- Implementing PR for #786: [#789](https://github.com/mbari-org/dash5/pull/789)

PR #789, opened 2026-07-18
fix: gate write actions behind login while keeping views open (#786)

### LINKED ISSUE #786: bug: gate write actions behind login (views stay open)
## Summary

Dash5 should remain **readable without login**, but **every write / edit path should be gated at entry** so users never start a mutating flow they cannot complete.

This issue started as a concrete document-edit failure reported by Monique, then expanded after auditing what login currently blocks (and what it does not).

**Status (paper trail for #789):** Must-fix / acceptance items below are done and verified against PR #789. Optional resume-after-login items deferred to #795. Overview map marker editing while logged out is tracked in #788.

---

## History / what brought us here

**Slack report (Monique Messié, 2026-07-17):**
> Could you please update dash5 such that it is not possible to edit a document (ie click the edit button) without being logged in? It has been a few times where I get caught off guard because I edit it, but then can't save the changes because I'm not logged in, and there is no way to do so without closing the window and losing all changes.

That is a real lost-work failure mode: the UI allows Edit while logged out, save fails (or is impossible), and there is no in-editor path to authenticate without discarding changes.

While investigating that report, we reviewed current auth gating in Dash5:

- Login is **not** required to browse the app. Overview, vehicles, map, schedule, logs, charts, docs (view), battery, ESP, and comms are all available while logged out.
- Many **mutating** modals *are* already wrapped in `requireAuthentication` in `Layout.tsx` (new/edit deployment, reassign, send note, new mission/command, attach/detach document, schedule event details, notifications, server status). Unauthenticated users who open those get the login modal instead.
- A few UI controls are already hidden when logged out (e.g. **Add Document**, **Add Note**, resources slot).
- Document **edit** and **add** modals are **not** wrapped in `requireAuthentication` — this is the Monique hole.
- The Docs ⋯ menu still offers Edit / Attach / Duplicate / Delete while logged out (attach/detach then bounce to login; edit does not).
- Schedule still shows mission/command affordances when logged out; opening those modals redirects to login (less painful than docs, but still inconsistent UX).

**Product decision:** keep **views open**; gate the **rest** (writes) at entry so users cannot enter an edit session they cannot save.

---

## Policy

| Open without login | Require login before the action starts |
|---|---|
| Overview / map / vehicle pages | Edit / add / duplicate / delete documents |
| Read-only docs, logs, charts, schedule, battery, ESP, comms | Attach / detach documents |
| Deployment / status viewing | Notes, missions, commands, deployment create/edit |
| | PIC / On-Call reassignment, notifications, server status |

**Why this policy**
- Viewing without login is valuable for operators/scientists who need situational awareness without friction.
- Write paths that fail only at save create the worst UX (lost work, confusion, no recovery path).
- Dash5 already mostly follows “actions gated, views open”; we should finish that pattern consistently instead of inventing a new “must login to view” model.

---

## Original bug (document edit) — still the highest priority

**To reproduce**
1. Open Dash5 while logged out (or with an expired session)
2. Open a document and click **Edit**
3. Make edits
4. Attempt to save — blocked / fails because the user is not logged in
5. No way to log in from the editor without closing and losing changes

**Expected (docs)**
- [x] Document Edit / Fill in is disabled or hidden when not logged in
- [x] Document Add / Duplicate / Delete entry points are likewise gated
- [x] Docs ⋯ menu does not offer write actions while logged out (or selecting them opens login immediately, without entering the editor)
- [x] Prefer clear affordance (tooltip / “Sign in to edit”) over a silent dead-end — *shipped as hidden write controls + existing toolbar Login (no extra banner/hint copy)*
- [ ] Optional nicety later: after login, return the user to the same document (view or edit intent) — **deferred**

---

## Suggested gates to finish the policy

Apply the existing `requireAuthentication` / hide-or-disable pattern consistently. Prefer **hiding or disabling write affordances in the UI**, not only swapping the modal for login after the user has already committed to an edit flow.

### Must-fix (causes lost work today)
- [x] **`editDocument` modal** — wrap in `requireAuthentication` *and* hide/disable Edit inside `DocumentInstanceModal` when unauthenticated
- [x] **`addDocument` modal** — same treatment (Add is hidden in DocsSection already, but the modal itself is not gated in Layout)
- [x] **Docs ⋯ menu write actions** while logged out: Edit, Duplicate, Delete (and Attach if still shown — Attach/Detach modals are already login-gated, but the menu item should not imply the action is available)

### Align remaining write entry points (consistency / trap prevention)
- [x] **Schedule** mission / command buttons — hide or disable when logged out (modals already redirect to login; avoid presenting a dead-looking or surprise-login click)
- [x] **Deployment create/edit** toolbar actions — ensure unauthenticated users see login (or disabled controls), not a partial form
- [x] **PIC / On-Call reassign** — already login-gated at modal; keep labels honest when logged out (`Unavailable` pattern is fine)
- [x] Any other write entry point discovered during implementation that lets a user start mutating UI without auth — *known remaining gap: overview map marker editing → #788*

### Explicitly out of scope (keep open)
- Viewing overview, map, vehicles, schedule, logs, charts, docs (read-only), battery, ESP, comms
- No app-wide “must be logged in to use Dash5” redirect

### Optional follow-ups (nice, not blocking) — deferred → #795
- [ ] After successful login from a gated action, resume the intended action (e.g. reopen the document for edit) — **deferred → #795**
- [ ] Allow login without discarding an open read-only document window — **deferred → #795**

---

## Implementation notes

- Primary existing pattern: `requireAuthentication` in `apps/lrauv-dash2/components/Layout.tsx` — used by most mutating modals; create/duplicate document flows are gated; existing-doc view stays open while logged out.
- Secondary pattern: hide buttons when `authenticated` is false (Docs / Schedule / deployment toolbar / `DocCell`).
- Best UX: combine both — don’t show write controls when logged out, and still gate mutating modals in Layout as defense in depth.

---

## Acceptance criteria

- [x] Logged-out users can still view Dash5 (map, vehicles, docs read-only, schedule, etc.)
- [x] Logged-out users cannot enter document edit/create/duplicate flows
- [x] Logged-out users cannot reach other write flows without being prompted to log in *before* making changes
- [x] No path where a user can accumulate unsaved edits and then discover they cannot save or log in without losing work

ksalamy:
Review Ready.

Review order: Copilot first, then human reviewers (requested when clean).

## Summary
- Gate Dash5 write/edit entry points so logged-out users cannot start mutating flows they cannot complete (fixes Monique’s lost-work doc-edit trap from #786).
- Keep browsing/viewing open without login (overview, map, vehicles, schedule, read-only docs, logs, charts, battery, ESP, comms).
- Logged-out UX relies on hidden write controls plus the existing toolbar Login (no signed-out banner or section “Sign in…” hint copy).

### Auth model (important nuance vs issue checklist)
- **Existing documents remain viewable while logged out.** We intentionally do *not* wrap all of `editDocument` in `requireAuthentication`, because that would block read-only viewing.
- **Create / duplicate** document flows *do* use `requireAuthentication` (they open in edit mode).
- **Edit / Fill in / Delete Revision** controls are hidden inside `DocumentInstanceModal` when unauthenticated.
- **Add Document** modal is gated in Layout; Add button already hidden when logged out.
- Docs ⋯ menu and detach controls are omitted when logged out (not just login-swapped after click). `DocCell` / OverviewToolbar write UI is driven by an explicit `authenticated` prop.

### Also gated for consistency
- Schedule: Mission / Command / Stop–Resume All; schedule ⋯ write actions (Download SBDs remains available).
- Deployment toolbar: “New … deployment” and Deployment Details hidden when logged out.
- Expanded Docs modal (`DocsModal`) now receives `authenticated` so logged-in users keep write controls there too.

### Out of scope / follow-ups
- Optional resume-after-login / login without closing a read-only doc window — follow-up: #795.
- Overview map marker editing while logged out — follow-up: #788.
- No app-wide “must be logged in to use Dash5” redirect.
- Admin/operator-style UI privileges (boolean `authenticated` only in this PR).

Closes #786

## Test plan
*(Will run and check off before merge.)*
- [x] Logged out: open a document → view OK, no Edit / Fill in / Delete Revision
- [x] Logged out: Docs section — no Add Document; no ⋯ menu; attachments visible as read-only labels (not detachable)
- [x] Logged out: Schedule — no Mission/Command/Stop All; ⋯ write actions hidden
- [x] Logged out: no New deployment / Deployment Details in vehicle toolbar
- [x] Logged out: no signed-out banner under the toolbar; toolbar Login still available
- [x] Logged in: accordion Docs and expanded Docs modal both show Add Document / ⋯ menu
- [x] Logged in: document Edit / save still works
- [x] Unit tests: DocsModal, DocsSection, DocumentInstanceModal, DocCell, OverviewToolbar, ScheduleSection


apps/lrauv-dash2/components/Layout.tsx
```diff
@@ -186,6 +224,34 @@ const Layout: React.FC<{ children?: React.ReactNode }> = ({ children }) => {
...
+              type="button"
+              onClick={dismissAuthBanner}
+              className="rounded px-2 py-1 text-xs text-stone-500 hover:bg-sky-100 hover:text-stone-700"
+              aria-label="Don't show this sign-in reminder again"
+            >
+              Don&apos;t show again
+            </button>
+          </div>
+        </div>
+      )}
```
ewiederhold: I know this is dismissible, but it's extremely redundant to the already existing Login button in the topmost navigation bar. I would strongly advise against adding this unless users are passionate about including a whole extra row warning them that they aren't logged in.

packages/react-ui/src/Cells/DocCell.tsx
```diff
@@ -15,8 +15,10 @@ export interface DocCellProps {
   className?: string
   style?: React.CSSProperties
   onSelect: () => void
-  onSelectAttachment: (attachment: Attachment) => void
-  onMoreClick: (
+  /** When omitted, attachments render as read-only labels (no detach control). */
+  onSelectAttachment?: (attachment: Attachment) => void
+  /** When omitted, the more-options menu is hidden. */
+  onMoreClick?: (
```
ewiederhold: since you're still always passing these arguments, I would type them as a union between the function that returns void and undefined (add `| undefined`) instead of making them optional, since you're not actually using them as optionals anywhere right now

packages/react-ui/src/Cells/DocCell.tsx
```diff
@@ -84,26 +86,34 @@ export const DocCell: React.FC<DocCellProps> = ({
...
+                    label={attachment.name}
+                    icon={faTimes as IconProp}
+                    onClick={swallow(() => onSelectAttachment(attachment))}
+                    reverse={true}
+                  />
+                ) : (
+                  <span className="mb-1 block text-sm font-semibold text-gray-700">
+                    {attachment.name}
+                  </span>
+                )}
```
ewiederhold: do we even want to show this list of attachments when someone isn't logged in? it doesn't seem very useful

apps/lrauv-dash2/components/DocsSection.tsx
```diff
@@ -187,6 +187,11 @@ const DocsSection: React.FC<DocsSectionProps> = ({
     <>
       <header className="flex p-2">
         <div className="flex flex-grow flex-col">
+          {!authenticated && (
+            <p className="mb-1 text-sm italic text-stone-600">
+              Sign in to add or edit
+            </p>
+          )}
```
ewiederhold: I'm again questioning the utility of adding more text; I think it will be pretty obvious to people who can normally add/edit documents that they need to sign in first when the buttons they expect to see aren't present and everything visually looks read-only

packages/react-ui/src/Cells/DocCell.tsx
```diff
@@ -84,26 +86,34 @@ export const DocCell: React.FC<DocCellProps> = ({
...
       </article>
-      <div className={styles.iconButton} ref={moreButtonRef}>
-        <IconButton
-          icon={faEllipsisV}
-          ariaLabel={'More options'}
-          onClick={handleMoreClick}
-          size={'text-2xl'}
-        />
-      </div>
+      {onMoreClick && (
```
ewiederhold: `onMoreClick` is a function definition, so it feels a bit weird to be using its existence (or lack of existence) to determine whether we should display certain parts of the UI – it feels like a state more like `authenticated` should just be passed down here instead

apps/lrauv-dash2/lib/shouldShowAuthBanner.ts
ewiederhold, line 1: this feels extremely overcomplicated for what could be a single line of evaluating `mounted && !authenticated && !authLoading && !dismissed` in the one place that `shouldShowAuthBanner` is utilized (in `Layout.tsx`) – do we really want to add all of this complexity to the codebase for a boolean check? I'd argue it doesn't need its own file and test and everything for a one liner

_maybe_ for the local storage constant, but this still feels over-engineered for what it's doing

apps/lrauv-dash2/components/VehicleAccordion.tsx
```diff
@@ -245,6 +245,7 @@ const VehicleAccordion: React.FC<VehicleAccordionProps> = ({
       />
       {section === 'schedule' && (
         <ScheduleSection
+          authenticated={!!authenticated}
```
ewiederhold: why is this one `!!authenticated` while every other passing of `authenticated` as a prop within `VehicleAccordion` isn't preceded with a double not?

packages/react-ui/src/Toolbars/OverviewToolbar.tsx
```diff
@@ -101,25 +101,30 @@ export const OverviewToolbar: React.FC<OverviewToolbarProps> = ({
...
-  const deploymentOptions =
-    deployments?.map((d) => ({
-      label: d.name,
-      onSelect: () => {
-        handleSelectDeployment?.(d)
-        setShowDeployments(false)
-      },
-    })) ?? []
+  const newDeploymentOptions =
+    authenticated && handleNewDeployment
```
ewiederhold: similar comment here as earlier in `DocCell.tsx`; I'm not convinced we should be using the existence or non-existence of functions to determine what we show on the UI (I would just use `authenticated` here)

ewiederhold (changes requested):
in addition to my other questions and comments throughout, I think the cruft in `Layout.tsx` should be cleaned up because with the new logic introduced in this PR, we're duplicate checking authentication status and the old method of doing that isn't cleaned up in this PR

other questions:

- who runs and reports on how the Test Plan went?
- do we attach output from running of unit tests to PRs?
- who goes back and checks whether we actually accomplished the todo items listed in the original ticket that will be automatically closed when this PR is merged? I see some checkboxes in #786 that haven't been accomplished in this PR and not sure where the paper train goes from here if everything gets closed
- how are we handling authentication of different privileges of users (admin, operator, ...)? I think there's more nuance than just the `authenticated` true/false state, but not sure where that happens or if we're even considering privilege within dash5 yet

happy to talk through any of this in person when you're back from vacation this week :)

apps/lrauv-dash2/components/Layout.tsx
ksalamy: Agreed — this **IS** a bit redundant. I will remove the signed-out banner entirely.

Logged-out users will still see write controls hidden and can use the existing toolbar Login.

packages/react-ui/src/Cells/DocCell.tsx
ksalamy: This is fair given the current “pass handler or `undefined`” call sites. I’ll address this together with the next point by driving UI from `authenticated` and keeping the handlers as real actions (always passed). That avoids the optional-vs-`| undefined` debate and matches where auth should go.

packages/react-ui/src/Cells/DocCell.tsx
ksalamy: I’d like to keep showing attachments while logged out, as read-only labels (no detach). They ARE a useful view context (which vehicle/deployment a doc is attached to). What’s gated is the mutating detach control, not the list itself. If that still feels noisy after you look at it, we can revisit — but my default is to keep the read-only list.

apps/lrauv-dash2/components/DocsSection.tsx
ksalamy: Ok. I’ll remove the hint. Same for the Schedule “Sign in to add missions or commands” line for consistency. I’ll keep the actual gating (hide Mission/Command/Add Document/⋯ write actions when logged out).

packages/react-ui/src/Cells/DocCell.tsx
ksalamy: Ok. I’ll switch this to an explicit `authenticated` prop on `DocCell` for show/hide of the ⋯ menu and detach controls. Handlers stay handlers (always passed from `DocsSection`); visibility comes from auth state — not from whether the function is defined.

apps/lrauv-dash2/lib/shouldShowAuthBanner.ts
ksalamy: I understand your concern regarding the code weight for a one-liner. This was for users who have commented in the past they want a clearer signed-out cue. Toolbar Login is enough for now, so I’ll remove the banner and delete `shouldShowAuthBanner.ts` + its test with it.

apps/lrauv-dash2/components/VehicleAccordion.tsx
ksalamy: The `!!` is only on Schedule because that section requires a real true/false, while the others still allow it to be missing. I’ll clean that up in `VehicleAccordion`, then pass the same value to Handoff, Schedule, and Docs — keeping the `authenticated` prop, with matching call sites.

packages/react-ui/src/Toolbars/OverviewToolbar.tsx
ksalamy: Same principle as `DocCell`. `authenticated` will drive the write affordances (e.g. “New … deployment”, Deployment Details), not whether handlers are wired or not. Handlers will stay optional callbacks for the actions themselves. I will also keep the empty-menu guard so the deployment toggle doesn’t open with no options when logged out.

ksalamy:
**Replying to your review notes on Layout cruft / process questions:**

Agreed. I’ll remove that unused state/effect in this PR.

I’ll keep the intentional auth model: `requireAuthentication` for mutating modals, plus threading `authenticated` into sections/toolbars so write entry points stay gated while views remain open. That’s the path we want to expand on later (not throw away).

**On your process questions:**

- **Test plan:** I’ll run this PR’s Test Plan before merge and check off the items in the PR description.

- **Unit-test output on PRs:** I don’t paste local logs. I use CI on the PR and fix failures before merge unless something is explicitly waived.

- **Issue 786 checkboxes / paper trail:** Fair point — `Closes` will shut Issue 786 even if its boxes are unchecked. Before merge I’ll tick what’s done, park the rest in follow-ups (map markers = Issue 788; resume-after-login stays deferred), then close.

- **Privileges (admin / operator / …):** Outside what this PR is trying to solve. This PR is gating on logged-in vs not (`authenticated`). Admin/operator-style UI privileges would be a follow-up if/when we need them; some role checks already live on the API (e.g. server health 403). I’m keeping the prop so we can extend that later without rewiring everything.

packages/react-ui/src/Cells/DocCell.tsx
ewiederhold: yeah I guess I just don't see the utility of being able to see those when you're not logged in, I don't feel strongly enough about it to request taking them out though (they are visual clutter, but I'd guess every person who looks at documents and cares about their attachments will always be logged in anyways)

confusingly though, the default mouse turns into a pointer when hovering over those labels, which makes me think they're clickable but then nothing happens when you click – that's not great UX and I've seen that in lots of places throughout the dash, I can make a ticket for future work

ksalamy:
Good Dog-food testing :-). Feel free to add a new issue to update the mouse on hover (I just missed that and forgot to update it with the Docs gate). You are welcome to assign yourself to it as well if you think that would be a quick second task to take on. 

As far as there being visual clutter still in the Docs, I am also not willing to die on this hill. It is probably best posed to a wider PIC user group. Feel free to request and see if anyone (including Monique) is that tied to having docs display even when not logged in. 

Thanks.

apps/lrauv-dash2/components/VehicleAccordion.tsx
ewiederhold: gotcha, do we want to assume that `authenticated` is `False` if it isn't passed (is `null`)? presumably `False` is a good default in this context, but I'm wondering whether that prop should just be required instead of optional to prevent weird edge case bugs

ewiederhold:
already opened #803 :)

new questions after reviewing again:
- why do we require login to view who's currently signed in as PIC, but not remove the `Handoff / On-Call` drawer when signed out? if we're protecting people's names, we should also remove that drawer when a user isn't logged in
- why are there two test files for `DocsSection` (one in /components & one in /jest)? they look very redundant
- in the `OverviewToolbar` component, the deployment dropdown is confusing both while logged in and not logged in; in both cases, the currently selected deployment is bolded at the top but the row is duplicated instead of having a single row per deployment (including the selected deployment) and just highlighting the selected deployment row instead of duplicating it – this visual confusion is made worse in the logged out view (especially when the selected deployment is, by default, the most recent one) because then there are two duplicate rows right on top of each other (see screenshot)
Screenshot 2026-07-31 at 12 19 36 PM

apps/lrauv-dash2/components/VehicleAccordion.tsx
ksalamy: The false default is intentional — fail-closed. `authenticated` is defined upstream in @mbari/api-client's useTethysApiContext() as `undefined | boolean` during initial load. Making it required **doesn't** really help — the upstream type doesn't change. Defaulting to false keeps writes gated until auth resolves. 

I can add a JSDoc comment to the prop definition to make the intent explicit if that helps.  Overall, this is a "later issue" (non-blocking) to this PR for you to discuss with Carlos.

ksalamy:
Saw the #803 after I posted the last comment. Thanks! :-)

On these two — both are pre-existing and out of scope for this PR, which was scoped to gating write actions only.

Showing PIC info when logged out is intentional read-only context (consistent with the attachment label decision). Happy to discuss with the wider team if there's appetite to gate read views too —> new issue if so.

The duplicate deployment row is a pre-existing bug in OverviewToolbar, not introduced here. Also worth its own issue (I checked and it does not have a dedicated issue yet if you wish to create it).

Can we keep this PR focused on the write-gate scope? Thanks.

ewiederhold:
yep, sorry, not requesting those as changes via this PR, just wanted to open the discussion somewhere relevant and have the digital paper trail :) I think talking to the team more broadly about what should/shouldn't be viewable when logged out would be great

for the duplicate row in the `OverviewToolbar` I made #804 

and then I did add one more comment to the above that I think is more relevant to this PR and should either be resolved via this PR or elsewhere:

> why are there two test files for DocsSection (one in /components & one in /jest)? they look very redundant

apps/lrauv-dash2/components/VehicleAccordion.tsx
ewiederhold: okay! what exactly do you mean to be discussed with Carlos though? just how we handle authentication between front- and back-end? is he responsible for everything in the `api-client` directory? I was under the impression those are wrappers for API calls, where he maintains the API itself, but we maintain the wrappers

ksalamy:
Good catch. Both files exist due to convention drift — Zack wrote the original alongside the component (Apr 2026), and I added a more complete version to /jest/ in PR #623 (May 2026). The /jest/ version is a strict superset, so I'll delete /components/DocsSection.test.tsx here. I found three other duplicate pairs (CommsSection, LogsSection, ScheduleSection) created similarly that will be consolidated in a follow-up PR, and I'll add a rule to search for previously created tests prior to writing new ones and enforce /jest/ going forward.


**Update:**  Deleted components/DocsSection.test.tsx and committed it to the PR #789 branch. A rule has been added to CLAUDE.md to enforce /jest/ going forward and to check for existing tests before writing new ones — this will be visible to all contributors.

apps/lrauv-dash2/components/VehicleAccordion.tsx
ksalamy: You're right — api-client **IS** ours, Carlos owns the TethysDash API itself. The "**discuss with Carlos**" was specifically about whether the backend should also enforce auth server-side (i.e., reject write commands from unauthenticated users at the API level), not just rely on the frontend hiding the controls. **That's a backend policy question**, not a wrapper question — hence Carlos. Happy to clarify that in a comment on the issue if useful.

**All agreed changes are in:**

1. Signed-out banner removed + shouldShowAuthBanner.ts and its test deleted
2. DocCell and OverviewToolbar now drive write UI from authenticated, not handler existence
3. Sign-in hint text removed from ScheduleSection
4. !!authenticated inconsistency resolved — const isAuthenticated = !!authenticated derived once at the top of VehicleAccordion and passed consistently to all three child sections (HandoffSection, ScheduleSection, DocsSection)
5. Redundant components/DocsSection.test.tsx deleted; /jest/ version is canonical
6. CLAUDE.md updated to enforce /jest/ going forward

Perhaps it is time to let this PR through? 😊

ewiederhold (approved):

PR #790, opened 2026-07-20
fix: gate map marker edits behind login

### LINKED ISSUE #788: follow-up: gate overview map marker editing behind login

ksalamy:
## Summary
- Keep overview/vehicle map markers visible when logged out; gate add/edit/delete/drag/recolor/save-to-layer behind authentication (same views-open / writes-gated rule as #786)
- Hide Add-markers control and MarkerSection rename/remove when logged out; `DraggableMarker` shows view-only popup without write handlers
- Add focused `DraggableMarker` tests for read-only vs editable popup controls

Closes #788

Related: #786, #789

## Test plan
- [ ] Logged out on Overview: markers remain visible; no Add-markers control; marker popup is view-only (label/position/Close only)
- [ ] Logged out: cannot drag markers; layers list keeps visibility checkbox but hides rename/remove
- [ ] Logged in: Add-markers, edit, delete, drag, recolor, save-to-layer work as before
- [ ] Logout while in add-marker mode clears add mode
- [ ] `yarn workspace @mbari/lrauv-dash2 test:ci components/DraggableMarker.test.tsx` passes

### DIFF
--- a/apps/lrauv-dash2/components/DraggableMarker.test.tsx
+++ b/apps/lrauv-dash2/components/DraggableMarker.test.tsx
@@ -93,4 +93,18 @@ describe('DraggableMarker write controls', () => {
     expect(screen.getByTitle('Save to Map Layers')).toBeInTheDocument()
     expect(screen.getByTitle('Close')).toBeInTheDocument()
   })
+
+  it('notifies parent when edit mode changes via onEditStateChange', () => {
+    const onEditStateChange = jest.fn()
+    render(
+      <DraggableMarker
+        {...baseProps}
+        onEdit={jest.fn()}
+        onEditStateChange={onEditStateChange}
+      />
+    )
+
+    // Effect notifies parent of current editMode (false on mount).
+    expect(onEditStateChange).toHaveBeenCalledWith(false)
+  })
 })
--- a/apps/lrauv-dash2/components/DraggableMarker.test.tsx
+++ b/apps/lrauv-dash2/components/DraggableMarker.test.tsx
@@ -1,6 +1,7 @@
 import '@testing-library/jest-dom'
 import React from 'react'
 import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
 
 jest.mock('react-leaflet', () => {
   const mockReact = require('react')
@@ -94,7 +95,8 @@ describe('DraggableMarker write controls', () => {
     expect(screen.getByTitle('Close')).toBeInTheDocument()
   })
 
-  it('notifies parent when edit mode changes via onEditStateChange', () => {
+  it('notifies parent when the user enters edit mode', async () => {
+    const user = userEvent.setup()
     const onEditStateChange = jest.fn()
     render(
       <DraggableMarker
@@ -104,7 +106,8 @@ describe('DraggableMarker write controls', () => {
       />
     )
 
-    // Effect notifies parent of current editMode (false on mount).
-    expect(onEditStateChange).toHaveBeenCalledWith(false)
+    onEditStateChange.mockClear()
+    await user.click(screen.getByTitle('Edit marker'))
+    expect(onEditStateChange).toHaveBeenCalledWith(true)
   })
 })
--- a/apps/lrauv-dash2/components/DraggableMarker.test.tsx
+++ b/apps/lrauv-dash2/components/DraggableMarker.test.tsx
@@ -1,0 +1,96 @@
+import '@testing-library/jest-dom'
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+
+jest.mock('react-leaflet', () => {
+  const mockReact = require('react')
+  return {
+    Marker: mockReact.forwardRef(
+      (
+        { children }: { children?: React.ReactNode },
+        _ref: React.Ref<HTMLDivElement>
+      ) => <div data-testid="leaflet-marker">{children}</div>
+    ),
+    Popup: mockReact.forwardRef(
+      (
+        { children }: { children?: React.ReactNode },
+        _ref: React.Ref<HTMLDivElement>
+      ) => <div data-testid="leaflet-popup">{children}</div>
+    ),
+    Tooltip: ({ children }: { children?: React.ReactNode }) => (
+      <div data-testid="leaflet-tooltip">{children}</div>
+    ),
+    useMap: () => ({}),
+  }
+})
+
+jest.mock('leaflet', () => ({
+  divIcon: jest.fn(() => ({})),
+}))
+
+jest.mock('react-dom/server', () => ({
+  renderToString: jest.fn(() => '<div />'),
+}))
+
+jest.mock('react-hot-toast', () => ({
+  __esModule: true,
+  default: Object.assign(jest.fn(), {
+    success: jest.fn(),
+    error: jest.fn(),
+  }),
+}))
+
+jest.mock('@mbari/utils', () => ({
+  createLogger: jest.fn(() => ({
+    debug: jest.fn(),
+    warn: jest.fn(),
+    error: jest.fn(),
+  })),
+}))
+
+jest.mock('./MarkerContext', () => ({
+  useMarkers: () => ({
+    handleMarkerSave: jest.fn(),
+    saveMarkerToLayer: jest.fn(),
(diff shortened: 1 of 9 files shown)


ksalamy:
@copilot review

apps/lrauv-dash2/lib/useClearMarkerEditModeWhenLoggedOut.ts
```diff
@@ -0,0 +1,41 @@
+import { useEffect } from 'react'
+
+/**
+ * Clears add/edit marker mode on logout and when the map view unmounts.
+ * MarkerProvider is app-wide, so edit state must not persist across routes.
```
ewiederhold, line 5: what does "edit state must not persist across routes" mean? if I place a marker in the `Overview` tab's map, and then switch to a vehicle-specific map (which updates the route), I still see the marker with the same edit menu as I saw on the `Overview` tab – if that's the type of persistence you're trying to prevent, I'm not sure it's working as intended

I think this also relates to the desire to clear state when a specific map view unmounts, but I'm not seeing that behavior when testing locally (although maybe I'm misunderstanding the intention there) – same as above, if I place a marker in the `Overview` tab and then navigate from `Map` to `Vehicles` and then back to `Map`, the marker and edit menu persists

apps/lrauv-dash2/lib/useClearMarkerEditModeWhenLoggedOut.ts
```diff
@@ -0,0 +1,41 @@
...
+    isAddingMarkers,
+    activeEditMarkerId,
+    setIsAddingMarkers,
+    setActiveEditMarkerId,
+  } = params
+
+  useEffect(() => {
+    if (!canEditMarkers) {
+      if (isAddingMarkers) setIsAddingMarkers(false)
+      if (activeEditMarkerId) setActiveEditMarkerId(null)
```
ewiederhold, line 25: can we set these two to `false` and `null` respectively, regardless of whether their associated state variable is true/truthy (eg no `if` statements needed here)? that would also eliminate two extra parameters needing to be passed to this hook

apps/lrauv-dash2/lib/getDraggableMarkerEditProps.ts
```diff
@@ -0,0 +1,35 @@
+type LatLng = [number, number]
+
+export type DraggableMarkerEditHandlers = {
+  onDragEnd: (pos: LatLng) => void
```
ewiederhold, line 4: why not inline this? it's only used here, and the `DraggableMarker` interface inlines it – it's also the same spelling as Leaflet's `LatLng` type which confused me at first
```suggestion
export type DraggableMarkerEditHandlers = {
  onDragEnd: (pos: [number, number]) => void
```

apps/lrauv-dash2/components/DraggableMarker.tsx
```diff
@@ -145,76 +146,103 @@ const DraggableMarker: React.FC<DraggableMarkerProps> = ({
...
 
-  // Handle marker click
-  const onEditStateChange = useCallback((isEditing: boolean) => {
-    // logger.debug(
-    //   `Marker edit state changed: ${isEditing ? 'Editing' : 'Not Editing'}`
-    // )
-  }, [])
+  const canEdit = !!onEdit
+  // Immediate view-only when handlers disappear (logout) — do not wait for
+  // the cleanup effect below to clear editMode after paint.
```
ewiederhold, line 151: what is "paint" in this context?

apps/lrauv-dash2/components/DraggableMarker.tsx
```diff
@@ -480,17 +510,29 @@ const DraggableMarker: React.FC<DraggableMarkerProps> = ({
...
+
+  const handleClosePopup = useCallback(
+    (e: React.MouseEvent) => {
+      e.stopPropagation()
+      e.preventDefault()
+      exitEditMode()
+      markerRef.current?.closePopup()
+    },
+    [exitEditMode]
+  )
```
ewiederhold, line 535: I think I understand why we need both of these, but the similar naming is really confusing and I'd suggest adding comments:
```suggestion
  // exit edit mode when the default Leaflet 'x' is clicked within the popup
  const handlePopupClose = useCallback(() => {
    exitEditMode()
  }, [exitEditMode])

  // additional cleanup when the custom blue 'x' (Close) button is clicked within the popup
  const handleClosePopup = useCallback(
    (e: React.MouseEvent) => {
      e.stopPropagation()
      e.preventDefault()
      exitEditMode()
      markerRef.current?.closePopup()
    },
    [exitEditMode]
  )
```

also more generally, I think that custom blue 'x' button should go away and we should just have one close button within the popup (being the default Leaflet one), I already captured that in #794 though

ewiederhold (commented):
few questions/comments/requests below :)

and then related to, but not needing to be fixed in this PR... the additions and deletions between `DeploymentMap.tsx` & `index.tsx` are completely identical – what are your thoughts about merging these map-related components into one so we don't have two places to update everything related to maps? maybe we should open a follow-up ticket for cleanup

apps/lrauv-dash2/components/DraggableMarker.test.tsx
ewiederhold, line 1: also one concrete request: I think we agreed to put new tests in `/jest`, not `/components`

ewiederhold (changes requested):

ksalamy:
**Testing notes (Sep 16, 2026)**

Verified on localhost against current `develop` + this branch.

- **Logged in:** Add, edit, delete, drag, recolor, and Save to Map Layers work on Overview and vehicle maps. Blue Save commits the pin on the map only; the green flag is what adds it to Layers. Session-only pins drop on refresh; layer-saved pins return.
- **Logged out:** Pin button is hidden. Marker popups are view-only. Markers stay visible. Layers keeps visibility checkboxes and hides Remove / Remove all from layer.
- **Logout mid-session:** Add mode and an open edit popup both clear. After login, adding another pin works. No stale `Please finish editing current marker first` toast.
- **Overview ↔ vehicle:** Pins remain (shared provider). Add/edit mode does not stay stuck. An unsaved new-marker popup no longer reopens on the next map.

ksalamy:
@copilot review

apps/lrauv-dash2/jest/DraggableMarker.test.tsx
ksalamy: I have moved this PRs `DraggableMarker.test.tsx` to the apps/lrauv-dash2/jest/ directory and have made a note that all new tests will go into the apps/lrauv-dash2/jest/ moving forward.

Issue #93, opened 2022-05-24
Implement `WaypointsTable` Component


jimjeffers:
This one might be a little trickier as we have to embed a select field in the table. I think this should still be doable via JSX. Let's not overthink it just yet. A proof of concept that embeds our `Fields.Select` component should do the trick but if you get blocked or have questions shoot them over and we'll figure it out! We can work on a specific select component for choosing waypoints after implementing this table.

## Scaffold the New Component

To scaffold this component generate a blank component:

```
yarn scaffold -t component ./packages/react-ui/src/Tables/WaypointsTable
```

## Reference Design

Use the following design in your story parameter:
https://www.figma.com/file/FtsKsOCBQ2YjTZlwezG6aI/MBARI-Components?node-id=3622%3A640

i.e.
```ts
Standard.parameters = {
  design: {
    type: "figma",
    url: "https://www.figma.com/file/FtsKsOCBQ2YjTZlwezG6aI/MBARI-Components?node-id=3622%3A640"
  }
}
```

PR #805, opened 2026-07-31
fix(#779): display mission ID in pending Schedule rows via getScript at queue time

### LINKED ISSUE #779: feat: display mission ID (not filename) consistently in Schedule tab
## Problem

Schedule tab shows filename for queued missions, switches to vehicle-reported mission ID once started — causing a confusing name change mid-execution.

## Goal

Show the correct mission ID from the moment a mission enters the queue.

## Implementation

Call `GET /TethysDash/api/commands/script-description?vehicle=...&scriptPath=...` at queue time to get the mission ID, then display it in pending Schedule rows.

Steps:
1. Add `getScriptDescription` to `packages/api-client/src/axios/Command/`
2. Substitute the returned mission ID for the filename in pending rows

> Carlos closed [mbari-org/tethysdash#93](https://github.com/mbari-org/tethysdash/issues/93) (won't fix) — POST /commands will not return the mission ID. `getScriptDescription` is the only path.

## References
- `script-description` endpoint: https://okeanids.mbari.org/TethysDash/q/swagger-ui/#/Commands/getScriptDescription
- Team discussion: see comments below

ksalamy:
> **Review Order:** Copilot review by author first, then human review will be requested.

## Problem
Pending Schedule rows showed the raw filename (e.g., `Transport/keepstation.tl` or `Science/profile_station_vt.tl`) instead of the mission ID (`keepstation`, `profile_station`). The correct ID was only visible after the vehicle started the mission and reported back via telemetry — causing a confusing mid-execution name change.

## Solution
At queue time, `ScheduleSection` now calls `GET /commands/script` for each pending mission (in parallel, with `staleTime: Infinity`) to resolve the mission ID immediately. Pending rows display the correct ID from the moment they enter the queue.

## Supersedes
This replaces the `_vt` regex approach from PR #774, which was merged and then reverted. That approach stripped `_vt` from filenames in the UI — an incorrect assumption. Quinn and Emme confirmed the right fix is to use the actual mission ID from `getScriptDescription`, not manipulate filenames. Carlos confirmed `GET /commands/script` is the correct and only available path (TethysDash #93 — POST /commands returning the ID is won't-fix).

## Testing
Verified locally on `sim` — `keepstation` and `profile_station` both show correct mission IDs as Pending before vehicle confirmation.

## References
- Closes #779
- Reverts #774

### DIFF
--- a/apps/lrauv-dash2/components/ScheduleSection.tsx
+++ b/apps/lrauv-dash2/components/ScheduleSection.tsx
@@ -29,6 +29,7 @@ import {
   useCreateNote,
   timeoutExpiredRegEx,
   useTethysApiContext,
+  getScript,
 } from '@mbari/api-client'
 import { useQueryClient, useQueries } from 'react-query'
 import useGlobalModalId from '../lib/useGlobalModalId'
@@ -40,7 +41,6 @@ import {
   normalizeMissionName,
   normalizeMissionPath,
 } from '../lib/missionUtils'
-import { getScript } from '@mbari/api-client'
 import { toast } from 'react-hot-toast'
 
 export interface ScheduleSectionProps {
--- a/apps/lrauv-dash2/components/ScheduleSection.tsx
+++ b/apps/lrauv-dash2/components/ScheduleSection.tsx
@@ -28,8 +28,9 @@ import {
   useDeleteCommandQueue,
   useCreateNote,
   timeoutExpiredRegEx,
+  useTethysApiContext,
 } from '@mbari/api-client'
-import { useQueryClient } from 'react-query'
+import { useQueryClient, useQueries } from 'react-query'
 import useGlobalModalId from '../lib/useGlobalModalId'
 import {
   missionNameFromStartedText,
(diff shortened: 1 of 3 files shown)


ksalamy:
**Note on `normalizeMissionName` and the `_vt` regex (PR #774):**

The `_vt` regex in `normalizeMissionName` is still present on `develop` and intentionally not removed here. It is used by `missionKeysMatch` for internal status tracking — correlating the file path in a command event with the vehicle-reported mission ID so rows transition correctly from `pending` → `running` → `completed`. Removing it without a replacement would break status tracking for `_vt` missions.

This PR addresses the **display** concern — pending rows now show the correct mission ID via `getScript` instead of the filename. The follow-up to properly remove the regex by using `getScript`-resolved IDs in the matching logic is tracked in #808.

ksalamy:
@copilot review

ksalamy:
**Testing notes:**

- All 44 tests pass, including ones that specifically check the mission name display before and after a row moves from active to history
- Checked manually/visually on localhost dev app: mission names look correct across all currently deployed vehicles (no file paths, no .tl extensions showing)
- Couldn't test a live _vt mission since none were running at the time, but that scenario is covered by the unit tests
- Also fixed a follow-up bug caught by Bugbot: once a mission name is looked up, it now sticks even if the row moves to history; previously it could revert back to the raw filename

Issue #796, opened 2026-07-28
bug: mission send uses filename for set commands instead of script mission ID


ksalamy:
## Summary

When scheduling a mission whose **filename** differs from the mission’s internal **ID**, Dash5 can build `set` commands with the wrong name. The vehicle then rejects the mission (syntax / mission-ID mismatch), even when the UI preview looks plausible.

Reported on Triton (Slack, 2026-07-28): Akshay sent `expanding_donut` via Dash5 from an earlier run; Quinn noted the vehicle mission ID is `expanding_donut` but Dash sent `tritoncam_expanding_donut`. Preview SBDs looked correct; the vehicle did not accept the mission. Sent with Dash5 (Dash4 retry suggested for comparison).

## Example

| | |
|---|---|
| Mission file | `Engineering/tritoncam_expanding_donut.tl` |
| Script / vehicle mission ID | `expanding_donut` |
| What Dash5 uses in `set …` | `tritoncam_expanding_donut` (filename stem) |

## Likely cause

`makeMissionCommand` derives the name for `set` from the file path:

```ts
const missionName = mission.split('/').pop()?.split('.')[0]
```

`GET /commands/script` already returns the real script `id`. That value should be used for `set` targets (and anywhere else the vehicle expects the mission ID), not the filename stem.

## Expected

- `load Engineering/tritoncam_expanding_donut.tl; …`
- `set expanding_donut.<param> …` (script mission ID)
- `run`

## Actual

- `set tritoncam_expanding_donut.<param> …` → vehicle rejects / stays in Default

## Related

- Related display work (not a substitute for this fix): #779 — Schedule tab showing filename vs mission ID
- Does **not** cover mission-picker listing / cache issues

## Repro sketch

1. Open Schedule → New Mission on a vehicle (e.g. Triton)
2. Select a mission where filename ≠ mission ID (e.g. `Engineering/tritoncam_expanding_donut.tl`)
3. Override at least one parameter and schedule / send
4. Compare generated `set` prefixes to the script `id` from `/commands/script`
5. Observe vehicle reject / remain in Default while Dash may still show ACK progress on SBDs

## Acceptance

- [ ] `set` commands use script mission `id`, not filename stem, when they differ
- [ ] Confirmed with a `tritoncam_*` (or similar) mission where filename ≠ ID
- [ ] Unit/integration coverage for `makeMissionCommand` (or send path) with mismatched filename vs `id`

PR #811, opened 2026-08-03
fix(#796): use script mission ID for set commands

ksalamy:
> **Review order:** Copilot review by author first, then human review will be requested.

`makeMissionCommand` used the `.tl` filename stem for both `load` and `set`. For missions where the script ID differs from the filename (e.g. `tritoncam_expanding_donut.tl` → ID `expanding_donut`), the `set` command was wrong.

`set` now prefers the `missionId` from `GET /commands/script`; `load` keeps the full file path. Falls back to filename stem when they match.

Verified on staging: `set profile_station.*` emitted correctly.


ewiederhold (approved):
looks good to me!

Issue #798, opened 2026-07-28
bug: Schedule shows Received/ACK before all SBD chunks arrive (cellsat multi-part)


ksalamy:
## Summary

On **cellsat** / mixed comms, Dash5 Schedule can show a mission as **Received** / ACK (“Received by &lt;vehicle&gt;”) while the vehicle has **not** yet received the full multi-SBD payload. Remaining chunks may still be sitting in the sat queue.

Reported by Emme Wiederhold / confirmed by Quinn Shemet (Slack, 2026-07-28) on Triton while waiting on `expanding_donut` (multi-SBD mission request, mixed cell+sat).

## Why this is misleading

Pilots read **Received / ACK** as “the vehicle has the command.” For long missions split across SBDs (e.g. `1/4` … `4/4`), that is only true when **all** chunks are delivered. Cell may succeed quickly (`sbdSend` `state === 2`) while sat is still chewing through later parts.

## Likely cause

In `determineCommandStatus`, any matching `sbdSend` with `state === 2` (cell) returns `status: 'ack'` immediately:

```ts
// Cell comms are considered ACKed when sent since an open socket means delivery.
if (matchingSbdSend?.state === 2 || via === undefined) {
  return { ... status: 'ack', ... }
}
```

There is no check that all SBD parts of a multi-fragment command have been received. Schedule then upgrades pending → ACK from that comms status (`ScheduleSection`).

This is **not** a regression of #584 (stuck pending after real ACK). That fix was correct for its case. This is a separate gap: **premature ACK** for cellsat + multi-part.

## Expected

- Do not show full **Received / ACK** for a multi-SBD command until delivery semantics match pilot expectation (all required chunks accounted for), **or**
- Keep status at **sent** / in-progress while parts remain outstanding, and only then ACK
- Ideally paired with Dash4-style chunk progress UI (#797) so waits are visible

## Actual

- Schedule shows ACK / “Received by Triton” while sat is still delivering remaining SBDs
- Operators think the vehicle has the full mission when it does not

## Acceptance

- [ ] cellsat multi-SBD mission does not show full Received/ACK until appropriate for complete delivery
- [ ] Single-chunk cell commands still ACK as today when that remains correct
- [ ] Timeout / #597 behavior unchanged (timeout note still wins over `state:2`)
- [ ] Tests covering cellsat multi-part premature ACK

## Related

- #797 — Dash4-style SBD chunk boxes (visibility; does not replace this status fix)
[...]

PR #812, opened 2026-08-03
fix(#798): do not ACK cellsat commands on cell state:2 alone

ksalamy:
> **Review order:** Copilot review by author first, then human review will be requested.

Schedule was showing Received/ACK when the cell leg completed even if the satellite leg was still delivering remaining SBD chunks. For `cellsat` commands, `state:2` (shore dispatched via cell) no longer counts as ACK — the command stays at `sent` until sat receipt/receive arrives (or times out). Pure `cell` ACK behavior is unchanged.

Verified passively on live vehicle Schedule rows.

### DIFF
--- a/packages/api-client/src/axios/Util/determineCommandStatus.test.ts
+++ b/packages/api-client/src/axios/Util/determineCommandStatus.test.ts
@@ -497,6 +497,103 @@ describe('determineCommandStatus', () => {
     expect(result.commsIsoTime).toBe(cellsatCellSend.isoTime)
   })
 
+  // #798: receipt without receive — still in flight on sat.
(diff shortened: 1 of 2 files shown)


ksalamy:
**Testing notes (Aug 20, 2026):** Validated via the restart logs command on pontus (Event ID 27542103, via:cellsat, timeout:5min). The row shows ACK (checkmark) with MTMSN: 2891 / MOMSN: 53724 present — confirming the ACK was triggered by Iridium sat delivery, not by cell state:2 alone. Two timed-out sysid_backseat rows from the same session correctly show timeout icons with no ACK, further confirming the fix. Bugbot and Copilot both clean.

